PR #8, opened 2017-02-19
feat: vars and reuse results

yesnault:

### DIFF
--- a/executors/http/executor.go
+++ b/executors/http/executor.go
@@ -54,7 +54,7 @@ func (Executor) GetDefaultAssertions() venom.StepAssertions {
 }
 
 // Run execute TestStep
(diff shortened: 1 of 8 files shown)


executors/http/executor.go
```diff
@@ -54,7 +54,7 @@ func (Executor) GetDefaultAssertions() venom.StepAssertions {
 }
 
 // Run execute TestStep
-func (Executor) Run(l *log.Entry, aliases venom.Aliases, step venom.TestStep) (venom.ExecutorResult, error) {
+func (Executor) Run(l *log.Entry, aliases venom.Aliases, step venom.TestStep, templater *venom.Templater) (venom.ExecutorResult, error) {
```
fsamin: maybe we should apply templater on all step params ?
it's a bit weird to have to handle this in all executor.

yesnault: yep, I'll fix

PR #10, opened 2017-02-20
feature: HTTP executor can POST multipart

geoffreybauduin:
I need this feature, but I find my method too "hacky" because of encode/decode and behaviors of the used encoding libraries.

**First issue I came up with, in templater:** 
```
templater> Error while marshaling: json: unsupported type: map[interface {}]interface {}
templater> Error while marshaling: json: unsupported type: map[interface {}]interface {}
templater> Error while marshaling: json: unsupported type: map[interface {}]interface {}
```

It appears `json` does not really like `map[interface{}]interface{}` type (and does not know what to do with it), while `yaml` (it's the one producing this type) is okay with it (hence the modification in `templater.go`)

**Second issue I came up with, in the executor**

Basically, `mapstructure.Decode` does not know how to handle the `MultipartForm` field. I had to set values manually (and assert types later)

If you have any ideas on how to make all this less "hacky", please don't hesitate, you probably use more Golang than I do.

Also, I am not sure how you're making sure any change is not breaking the software, so I did not look into testing.

### DIFF
--- a/executors/http/executor.go
+++ b/executors/http/executor.go
@@ -4,8 +4,13 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"io"
 	"io/ioutil"
+	"mime/multipart"
 	"net/http"
+	"os"
+	"path/filepath"
+	"strings"
 	"time"
 
 	log "github.com/Sirupsen/logrus"
@@ -28,11 +33,12 @@ type Headers map[string]string
 
 // Executor struct
 type Executor struct {
-	Method  string  `json:"method" yaml:"method"`
-	URL     string  `json:"url" yaml:"url"`
-	Path    string  `json:"path" yaml:"path"`
-	Body    string  `json:"body" yaml:"body"`
-	Headers Headers `json:"headers" yaml:"headers"`
+	Method        string  	  `json:"method" yaml:"method"`
+	URL           string  	  `json:"url" yaml:"url"`
+	Path          string  	  `json:"path" yaml:"path"`
+	Body          string  	  `json:"body" yaml:"body"`
+	MultipartForm interface{} `json:"multipart_form" yaml:"multipart_form"`
+	Headers       Headers 	  `json:"headers" yaml:"headers"`
 }
 
 // Result represents a step result
@@ -62,16 +68,12 @@ func (Executor) Run(l *log.Entry, aliases venom.Aliases, step venom.TestStep) (v
 		return nil, err
 	}
 
+	// dirty: mapstructure doesn't like decoding map[interface{}]interface{}, let's force manually
+	t.MultipartForm = step["multipart_form"]
+
 	r := Result{Executor: t}
 
-	body := bytes.NewBuffer([]byte(t.Body))
-
-	path := t.URL + t.Path
-	method := t.Method
-	if method == "" {
-		method = "GET"
-	}
-	req, err := http.NewRequest(method, path, body)
+	req, err := t.getRequest()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


yesnault:
\i/ Seems good, two points:
 - can you add an example in readme ? (or add a new Readme file like https://github.com/runabove/venom/pull/11/files#diff-79519ffc6c2ccac4e7862584694ee6e0 , but for exec executor) ?
 - can you one-line all err when it's possible ?

cc @fsamin @sguiheux

executors/http/executor.go
```diff
@@ -119,3 +121,71 @@ func (Executor) Run(l *log.Entry, aliases venom.Aliases, step venom.TestStep) (v
...
+	}
+	if e.Body != "" && e.MultipartForm != nil {
+		return nil, fmt.Errorf("Cannot use both 'body' and 'multipart_form'")
+	}
+	var body *bytes.Buffer
+	var writer *multipart.Writer
+	if e.Body != "" {
+		body = bytes.NewBuffer([]byte(e.Body))
+	} else if e.MultipartForm != nil {
+		form := e.MultipartForm.(map[interface{}]interface{})
```
fsamin: you should check if e.MultipartForm can not be cast to (map[interface{}]interface{})

executors/http/executor.go
```diff
@@ -119,3 +121,71 @@ func (Executor) Run(l *log.Entry, aliases venom.Aliases, step venom.TestStep) (v
...
+		body = &bytes.Buffer{}
+		writer = multipart.NewWriter(body)
+		for key_, value_ := range form {
+			key := key_.(string)
```
fsamin: may be you should try to cast e.MultipartForm to (map[string]string) directly to avoid cas of keys and values.
Or you must check `key, ok := key_.(string)`

executors/http/executor.go
```diff
@@ -119,3 +121,71 @@ func (Executor) Run(l *log.Entry, aliases venom.Aliases, step venom.TestStep) (v
...
+func (e Executor) getRequest() (*http.Request, error) {
+	path := fmt.Sprintf("%s%s", e.URL, e.Path)
+	method := e.Method
+	if method == "" {
+		method = "GET"
+	}
+	if e.Body != "" && e.MultipartForm != nil {
+		return nil, fmt.Errorf("Cannot use both 'body' and 'multipart_form'")
+	}
+	var body *bytes.Buffer
```
fsamin: body := &bytes.Buffer{} and remove the else one line 168

executors/http/executor.go
```diff
@@ -119,3 +121,71 @@ func (Executor) Run(l *log.Entry, aliases venom.Aliases, step venom.TestStep) (v
...
+	if e.Body != "" {
+		body = bytes.NewBuffer([]byte(e.Body))
+	} else if e.MultipartForm != nil {
+		form := e.MultipartForm.(map[interface{}]interface{})
+		body = &bytes.Buffer{}
+		writer = multipart.NewWriter(body)
+		for key_, value_ := range form {
+			key := key_.(string)
+			value := value_.(string)
+			if strings.HasPrefix(value, "@") {
```
fsamin: ?

executors/http/executor.go
```diff
@@ -119,3 +121,71 @@ func (Executor) Run(l *log.Entry, aliases venom.Aliases, step venom.TestStep) (v
...
+				// todo: how can we be sure the @ is not the value we wanted to use ?
+				_, err := os.Stat(value[1:])
+				if !os.IsNotExist(err) {
```
fsamin: if _, err := os.Stat(value[1:]); !os.IsNotExist(err) {

executors/http/executor.go
```diff
@@ -119,3 +121,71 @@ func (Executor) Run(l *log.Entry, aliases venom.Aliases, step venom.TestStep) (v
...
+					part, err := writer.CreateFormFile(key, filepath.Base(value[1:]))
+					if err != nil {
+						return nil, err
+					}
+					err = writeFile(part, value[1:])
```
fsamin: one line please

executors/http/executor.go
```diff
@@ -119,3 +121,71 @@ func (Executor) Run(l *log.Entry, aliases venom.Aliases, step venom.TestStep) (v
...
+					if err != nil {
+						return nil, err
+					}
+					continue
+				}
+			}
+			writer.WriteField(key, value)
```
fsamin: missing error checking i guess

geoffreybauduin: You're right, my IDE did not warn me about that. Will fix

executors/http/executor.go
geoffreybauduin: I'm getting inspired from curl which loads a file if data is prefixed by "@". Will document that

geoffreybauduin:
I considered all your remarks @fsamin

fsamin:
nice ! thank you very much

PR #17, opened 2017-02-21
feat: update & release system

### LINKED ISSUE #16: venum update
Update system / venom version

yesnault:
close #16

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: go
 go:
-- 1.8
+- tip
 
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,33 @@
+language: go
+go:
+- 1.8
+
+env:
+  global:
+  - PROJECT_NAME="venom"
+  - PROJECT_PATH="github.com/runabove"
+
+install: true
+
+script: chmod +x package.sh && ./package.sh
+
+before_deploy:
+  - gem install mime-types -v 2.6.2
+
+deploy:
+  provider: releases
+  api_key:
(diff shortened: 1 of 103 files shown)


venom.go
```diff
@@ -2,8 +2,14 @@ package venom
 
 import (
 	"fmt"
+	"os"
 )
 
+// Version of Blinkstick
```
fsamin: wrong copy & paste

venom.go
```diff
@@ -2,8 +2,14 @@ package venom
...
+// One Line for this, used by release.sh script
+// Keep "const Version on one line"
+const Version = "0.9.0"
```
fsamin: 0.1.0 ? or maybe it should be set on compilation

yesnault: 0.9.0 -> first release will be "1.0.0", by executing ./release.sh 1.0.0

.travis.yml
```diff
@@ -0,0 +1,33 @@
+language: go
+go:
+- 1.8
```
fsamin: you should set to `tip`

venom.go
fsamin: but why set 0.9 ? i would prefer "snapshot", if it's replaced on build time with tag name.

the very first public stable (1.0.0) will be released when we will considered it as stable - from a functionnal and design point of view (with plugin support for instance).

fsamin:
my last comment isn't displayed as well
https://github.com/runabove/venom/pull/17#discussion_r102339954

fsamin:
about https://github.com/runabove/venom/pull/17#pullrequestreview-23091926 ?

PR #41, opened 2017-02-26
feat(venom): add --exclude and check yml ext

yesnault:


process_files.go
```diff
@@ -5,25 +5,53 @@ import (
...
+		fpaths, errg := filepath.Glob(p)
 		if errg != nil {
 			log.Fatalf("Error reading files on path:%s :%s", path, errg)
 		}
-		filesPath = append(filesPath, fp...)
+		for _, fp := range fpaths {
+			toExclude := false
+			for _, te := range fpathsExcluded {
+				if te == fp {
+					toExclude = true
```
fsamin: break

process_files.go
```diff
@@ -5,25 +5,53 @@ import (
...
+				}
+			}
+			if toExclude {
+				log.Debugf("%s is excluded ", fp)
+			} else if strings.HasSuffix(fp, ".yml") || strings.HasSuffix(fp, ".yaml") {
+				filesPath = append(filesPath, fp)
+			} else {
+				log.Debugf("%s is skipped (not yaml extension)", fp)
```
fsamin: This log is useless.
I think i prefer a global log listing of files which will be computed by venom

process_files.go
yesnault: done

PR #42, opened 2017-02-27
feat(venom): extract named variables from result

yesnault:

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,27 @@ testcases:
 
 ```
 
+Extract variable from results and reuse it in step after
+
+```
+name: MyTestSuite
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -109,6 +109,27 @@ testcases:
...
+
+```
+name: MyTestSuite
+testcases:
+- name: testA
+  steps:
+  - type: exec
+    script: echo 'foo with a bar here'
+    extracts:
+      result.systemout: foo with a (?P<myvariable>[a-z]+) here
```
fsamin: {{myvariable=[a-z]+}}

PR #47, opened 2017-03-13
Use file for request body

WinterNis:
Venom will first try to load config from body if specified, then from bodyFile if specified.

### DIFF
--- a/executors/http/executor.go
+++ b/executors/http/executor.go
@@ -37,6 +37,7 @@ type Executor struct {
 	URL           string      `json:"url" yaml:"url"`
 	Path          string      `json:"path" yaml:"path"`
 	Body          string      `json:"body" yaml:"body"`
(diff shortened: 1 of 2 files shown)


executors/http/executor.go
```diff
@@ -129,14 +130,23 @@ func (e Executor) getRequest() (*http.Request, error) {
...
 		body = bytes.NewBuffer([]byte(e.Body))
-	} else if e.MultipartForm != nil {
+	} else if e.BodyFile != "" {
+        path := string(e.BodyFile)
+        if _, err := os.Stat(path); !os.IsNotExist(err) {
+            temp, err := ioutil.ReadFile(path)
+            if err != nil {
+                return nil, err
+            }
+            body = bytes.NewBuffer([]byte(temp))
```
fsamin: temp is already a byte array, no need to `[]byte(...)` just `bytes.NewBuffer(temp)`

Issue #5, opened 2017-02-18
feat: usability in unit tests


fsamin:
Would like to use venom steps in a golang unit tests as following

````golang
func TestSomething(t *testing.T) {
    steps := []venom.Step{
        {
             Type: "http",
             Method: "POST",
             URL: "https://eu.api.ovh.com/1.0/auth/logout",
             Assertions: []venom.Assertion{
                 {"result.statuscode", venom.Asserts.ShouldEqual, 401},
             },
        },
    }
    venom.Run(t, steps)
}
````
or 

````golang
func TestSomething(t *testing.T) {
    steps := []venom.Step{
        {
             Type: "http",
             Method: "POST",
             URL: "https://eu.api.ovh.com/1.0/auth/logout", //No default assertions at all
        },
    }
    v := venom.Run(t, steps)
    assert.NoError(t, v.Err())
    assert.NoEqual(t, v.Result["statuscode"], 401)
}
````

PR #52, opened 2017-04-09
feat: venom as Lib

fsamin:
Sample usage

```golang
func TestVenomTestCase(t *testing.T) {
	v := venom.TestCase(t, "TestVenomTestCase", venom.V{
		"foo": "bar",
	})
	r := v.Do(venom.H{
		"type":   "exec",
		"script": "echo {{.foo}}",
	})
	assert.Equal(t, "bar", r["result.systemout"])
}
```

still have to improve it; with some docs also

### DIFF
--- a/executors/imap/decode.go
+++ b/executors/imap/decode.go
@@ -12,8 +12,9 @@ import (
 	"net/mail"
 
 	log "github.com/Sirupsen/logrus"
-	"github.com/runabove/venom"
 	"github.com/yesnault/go-imap/imap"
+
+	"github.com/runabove/venom"
 )
 
 func decodeHeader(msg *mail.Message, headerName string) (string, error) {
--- a/executors/imap/decode.go
+++ b/executors/imap/decode.go
@@ -12,6 +12,7 @@ import (
 	"net/mail"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/runabove/venom"
 	"github.com/yesnault/go-imap/imap"
 )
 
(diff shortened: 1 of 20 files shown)


yesnault:
close #23

yesnault:
close #5

executors/imap/decode.go
```diff
@@ -12,6 +12,7 @@ import (
 	"net/mail"
 
 	log "github.com/Sirupsen/logrus"
+	"github.com/runabove/venom"
```
sguiheux: wrong place

lib/lib.go
```diff
@@ -0,0 +1,130 @@
...
+			Name:      name,
+		},
+		&venom.TestCase{
+			Name: name,
+		},
+		name,
+	}
+}
+
+//Do execuutes a veom test steps
```
sguiheux: only one u

lib/lib.go
```diff
@@ -0,0 +1,130 @@
...
+	venom.RegisterExecutor(readfile.Name, readfile.New())
+	venom.RegisterExecutor(smtp.Name, smtp.New())
+	venom.RegisterExecutor(ssh.Name, ssh.New())
+	venom.RegisterExecutor(web.Name, web.New())
+	venom.RegisterTestCaseContext(defaultctx.Name, defaultctx.New())
+	venom.RegisterTestCaseContext(webctx.Name, webctx.New())
+}
+
+//H is a map of test parameters
+type H map[string]interface{}
```
sguiheux: P

lib/lib.go
```diff
@@ -0,0 +1,130 @@
...
+		},
+		&venom.TestCase{
+			Name: name,
+		},
+		name,
+	}
+}
+
+//Do execuutes a veom test steps
+func (t *T) Do(teststep H) R {
```
sguiheux: testStepParam

PR #56, opened 2017-06-20
fix (executor/http): allow only PATH

yesnault:


yesnault:
allow all env.

PR #87, opened 2017-10-12
feat (assertion): support quoted args when splitting assertion string.

wI2L:
live example: https://play.golang.org/p/7X9XRj-I0I

### DIFF
--- a/assertion.go
+++ b/assertion.go
@@ -3,6 +3,8 @@ package venom
 import (
 	"fmt"
 	"strings"
+	"unicode"
+	"unicode/utf8"
 
 	"github.com/fsamin/go-dump"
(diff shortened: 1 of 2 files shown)


assertion.go
```diff
@@ -78,7 +80,8 @@ func applyAssertions(executorResult ExecutorResult, step TestStep, defaultAssert
 }
 
 func check(assertion string, executorResult ExecutorResult, l Logger) (*Failure, *Failure) {
-	assert := strings.Split(assertion, " ")
+	// assert := strings.Split(assertion, " ")
```
yesnault: remove comment

yesnault: Can you add some unit test on splitArgs ?

PR #91, opened 2017-10-16
fix: convert executor result to boolean for ShouldBeTrue/False assertions.

wI2L:


wI2L:
Should be fixed with https://github.com/ovh/venom/issues/92

PR #96, opened 2017-10-23
fix: replace ExecutorResult type to fix assertions.

wI2L:
Assertions are expected to use one or more `interface{}` arguments that hold a concrete type.
The breaking changes introduces by @fsamin in his package go-dump allow us to retrieve a `map[string]interface{}` dump of an executor result.

We use the concrete type of each field to properly convert the arguments(s) that are extracted from the assertions string. Support for  all range of `int` and `uint`, `bool`, `string`, `float32`, `floa64`, `time.Time`, `time.Duration`.


wI2L:
@yesnault @fsamin Since the package go-dump can actually dump anything, wouldn't it be simpler to return an interface (that ies expected to be a struct), and do the dumping to an ExecutorResult later on, transparentely, so the assertion process is abstracted ?

PR #106, opened 2017-11-09
feat(venom): pretty print XML and JSON output of testsuite results.

wI2L:
@yesnault @fsamin @sguiheux Should I add an option to disable pretty printing ?

fix https://github.com/ovh/venom/issues/94


wI2L:
Closes https://github.com/ovh/venom/issues/94

PR #114, opened 2017-11-26
feat: ignore some variables in parse

fsamin:


yesnault:
Yes, we need it for IT Test of CDS.

```
VENOM - try fo find var cds.build.cdsro in cds variables
VENOM - var cds.build.cdsro is found with value cds -w -f ~/.cds/it.ro.json
VENOM - try fo find var cds.build.cds in cds variables
VENOM - var cds.build.cds is found with value cds -w -f ~/.cds/it.rw.json
VENOM - try fo find var cds.build.cdsctl in cds variables
VENOM - var cds.build.cdsctl is found with value cdsctl
VENOM - Fail on venom: Missing variables [cds.version : Coucou name cds.proj.VAR_2 cds.version git.branch cds.build.foo cds.pipelinecds.stagecds.job cds.version. Coucou name gitURL in gitTarget; echo toto=toto goGetTarget; echo toto=toto goTestTarget; echo toto=toto cds.app.foo cds.env.yolo cds.proj.foo cds.pip.foo cds.buildNumber and vcds.version cds.buildNumber and vcds.version and env name is 'cds.env.name]
Starting step /plugin-venom-7
Starting plugin: plugin-venom version 74
End of step /plugin-venom-7 [Fail]
```

Hot fix on cds plugin -> https://github.com/ovh/cds/commit/cceb5446ddf5f8152a018f190662545f1eca2cab

IT Test CDS uses {{.cds.. }}  and venom really need to don't check them. It's normal that there is no replace for them.

process.go
```diff
@@ -24,48 +24,34 @@ func (v *Venom) Process(path []string, exclude []string) (*Tests, error) {
...
-	return testsResult, nil
-}
-
-func (v *Venom) parse() error {
 	missingVars := []string{}
 	extractedVars := []string{}
 	for i := range v.testsuites {
 		ts := &v.testsuites[i]
-		log.Info("Parsing testsuite %s", ts.Package)
+		log.Info("Parsing testsuite", ts.Package)
```
yesnault: log.Infof("Parsing testsuite %s", ts.Package)

process.go
```diff
@@ -106,7 +92,15 @@ func (v *Venom) parse() error {
 			}
 		}
 		if !varExtracted {
-			reallyMissingVars = append(reallyMissingVars, k)
+			var ignored bool
+			for _, i := range v.IgnoreVariables {
```
yesnault: where venom put something inside IgnoreVariables ?

process_files.go
```diff
@@ -30,19 +30,15 @@ func getFilesPath(path []string, exclude []string) ([]string, error) {
 	for _, p := range path {
 		p = strings.TrimSpace(p)
 
-		fileInfo, err := os.Stat(p)
-		if err != nil {
-			return nil, err
-		}
-
+		fileInfo, _ := os.Stat(p)
```
yesnault: A comment why we ignore err can be useful:
// no need to check err on os.stat.
// if we put ./test/*.yml, it will fail and it's normal

process.go
yesnault: Ok, in plugin-venom only.

fsamin: yep, but if you want to add flags on command line, i can do it. 
I had a global flag to avoid parse instead

process.go
fsamin: when the `%s` is at the end it's no very usefull in fact
https://play.golang.org/p/aUQ6D9vKy2

process_files.go
fsamin: done

PR #115, opened 2017-11-26
fix: no need to check err on os.stat

yesnault:
this will avoid that:

```
VENOM - try fo find var cds.build.cdsro in cds variables
VENOM - var cds.build.cdsro is found with value cds -w -f ~/.cds/it.ro.json
VENOM - try fo find var cds.build.cds in cds variables
VENOM - var cds.build.cds is found with value cds -w -f ~/.cds/it.rw.json
VENOM - try fo find var cds.build.cdsctl in cds variables
VENOM - var cds.build.cdsctl is found with value cdsctl
VENOM - Fail on venom: stat ./tests/*.yml: no such file or directory
Starting step /plugin-venom-7
Starting plugin: plugin-venom version 73
End of step /plugin-venom-7 [Fail]
```

### DIFF
--- a/process_files.go
+++ b/process_files.go
@@ -30,10 +30,9 @@ func getFilesPath(path []string, exclude []string) ([]string, error) {
 	for _, p := range path {
 		p = strings.TrimSpace(p)
 
-		fileInfo, err := os.Stat(p)
-		if err != nil {
-			return nil, err
-		}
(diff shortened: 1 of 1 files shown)


yesnault:
I close this, it's in https://github.com/ovh/venom/pull/114

PR #117, opened 2017-11-29
Update the output from `venom template` to match reality.

sean-:
Without this fix, the sample template produced by `venom template` creates an error and creates a bad taste in someone's mouth who is new to `venom`.


fsamin:
thanks @sean-

PR #120, opened 2017-12-01
feat: run testsuites in parallel

fsamin:


sean-:
Thank you for posting the parallel test suites patch.

Would you object if I added a dependency graph to venom for test cases and then added a second configuration option for the parallelism of test cases?

```
--parallel-test-cases=1 # default
--parallel-test-suites=1 # default, formerly known as --parallel
```

Inside of a test case, give each test case a name, and then let test cases build their own dependency graph using `before` or `after`.  This way some tests could be run concurrently.

My use case is that some test suites have very slow and expensive setup procedures, but once setup, the test cases are reasonably cheap and can be run concurrently.  The problem is I need to create a dependency graph between each of the test cases.  Test cases that all have the same prerequisite can be run in parallel.

I'm mostly done, actually, but wanted to catch someone with this before they merged this feature.  ?

fsamin:
Ability to run testcases in parallel inside testsuites sounds good to me, feel free to contribute.

PR #124, opened 2017-12-19
update go-ovh

marcaudefroy:
feat: update go-ovh (adding headers feature)
feat: add headers to ovh executor
feat: add insecure_tls to ovh executor

Signed-off-by: Marc Audefroy <marc.audefroy@corp.ovh.com>

### DIFF
--- a/executors/ovhapi/executor.go
+++ b/executors/ovhapi/executor.go
@@ -1,9 +1,11 @@
 package ovhapi
 
 import (
+	"crypto/tls"
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
+	"net/http"
 	"os"
 	"time"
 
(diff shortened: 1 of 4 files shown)


executors/ovhapi/executor.go
```diff
@@ -110,16 +112,38 @@ func (Executor) Run(testCaseContext venom.TestCaseContext, l venom.Logger, step
...
+	}
+
+	if e.Headers != nil {
+		for key := range e.Headers {
+			req.Header.Add(key, e.Headers[key])
+		}
+	}
+
 	// do api call
-	resp := new(interface{})
```
wI2L: Why did you renamed it ? Idiomatic naming is `resp`, that is what the standard library is using in packahe `net/http`.

PR #127, opened 2018-01-11
Migrate from govendor to dep

sean-:
Remove `govendor` and replace with `dep`.

Individual packages were updated using the rough procedure:

1. `dep ensure`
2. `dep ensure -update ${PACKAGE}`
3. `dep prune`
4. `go build`


yesnault:
LGTM

sean-:
Before I forget (or for the record), I deliberately installed either `[[constraint]]` or `[[override]]` for tables in `Gopkg.toml` for packages that had previously tagged releases, but not recently (or recent enough). Notably there were a few cases where tags were quite old and had drifted from `master` or where `venom` is using functionality that wasn't released. When using `dep ensure && dep ensure -update && dep prune`, it may become obvious in the future where various packages can become un-pinned in `Gopkg.toml`.

PR #128, opened 2018-01-11
Parse tests as HCL...

sean-:
This PR passes all tests.  I'd like it if someone at OVH could pull this branch down and run it through your paces locally to see if it breaks anything on your end.  I'm worried about one and only one change in this PR:

https://github.com/ovh/venom/commit/8bebffe2582d1f9e5623b3af950bb99eddb72b4c#diff-e871404fd69cab40521d6b2178847cf7R120

I think all of the decoders are doing the right thing, however, which is why that change was a nop and everything has continued to work as expected.  In particular I'm worried that the output of XML is somehow invalid or different. ?

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -90,6 +90,12 @@
   revision = "02f7e9439689eda7ce4752c7f37e50cb678f2679"
 
 [[projects]]
+  branch = "master"
+  name = "github.com/pkg/errors"
+  packages = ["."]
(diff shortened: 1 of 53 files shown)


yesnault:
@sean- assertion on TestCase seems not used

sean-:
@yesnault if that's the case, do you mind merging this? :grinning:

yesnault:
@fsamin ok for you too ?

fsamin:
👍

PR #142, opened 2018-03-20
Allow to have numbers not unmarshaled as float

akuracy:


yesnault:
This will be done on go-ovh, after a new release (go mod).

PR #145, opened 2018-04-11
add workdir on step to allow relative path into test files

marcaudefroy:
Currently, when venom runs tests placed into subdirectory, the relative path into files test isn't well used. Indeed, executors aren't access to workdir about a test file.  This PR fix this issue.


marcaudefroy:
Waiting before merge...

executors/dbfixtures/executor.go
```diff
@@ -62,9 +64,10 @@ func (e Executor) Run(testCaseContext venom.TestCaseContext, l venom.Logger, ste
 	// Load and import the schemas in the database
 	// if the argument is specified.
 	if len(e.Schemas) != 0 {
+
```
wI2L: ??

executors/dbfixtures/executor.go
```diff
@@ -42,7 +43,8 @@ type Result struct {
 }
 
 // Run implements the venom.Executor interface for Executor.
-func (e Executor) Run(testCaseContext venom.TestCaseContext, l venom.Logger, step venom.TestStep) (venom.ExecutorResult, error) {
+func (e Executor) Run(testCaseContext venom.TestCaseContext, l venom.Logger, step venom.TestStep, workdir string) (venom.ExecutorResult, error) {
+
```
wI2L: ??

marcaudefroy: PR is in progress

yesnault:
This will break existing users who use relative import in their yml files

Proposal: add optional version in yml files.

if version:2 -> use workdir, if no version (=version:1 -> use behaviour as today)

This will avoid to break test on existing users, but let you to add this cool feature in yml version 2.

marcaudefroy:
Yvonnick : version added.
To test it, there are two file into dbfixture. A file on version 1 (no version on testsuite) and a file on version 2.

process_files.go
```diff
@@ -107,6 +107,18 @@ func (v *Venom) readFiles(filesPath []string) (err error) {
 
 		ts.Name += " [" + f + "]"
 
+		if ts.Version == "2" {
```
sguiheux: if not startwith 1 will be better and will avoid to modify this line for the next version

PR #175, opened 2018-11-13
feat: http Executor can be configured as "no follow redirect"

rbeuque74:
If you want to test HTTP pages that returns 302 Redirect, this could be useful :)


rbeuque74:
Done

PR #179, opened 2018-12-13
Add support for talking HTTP over Unix socket

riton:
This P.R adds a new parameter to the `http` executor that allows to use the _Unix sockets_ transport layer.

Example:

```yaml
- name: GET API health over Unix Socket
  steps:
  - type: http
    method: GET
    unix_sock: /run/myapp/health.sock
    url: http://unix/health
    assertions:
    - result.bodyjson.success ShouldBeTrue
```


yesnault:
All is ok, thank you!

PR #206, opened 2019-10-16
fix: raise error on parsing

fsamin:


process_files.go
```diff
@@ -89,6 +89,7 @@ func (v *Venom) readFiles(filesPath []string) (err error) {
 			out = tmp
 		}
 
+		err = nil
```
golangcibot: ineffectual assignment to `err` (from `ineffassign`)

PR #218, opened 2020-01-20
fix(executor/http): support setting host in headers

### LINKED ISSUE #217: bug(httpExecutor): support changing request hostname
When building a http request in Go, the Host header isn't honored. From the docs:

~~~
	// For client requests Host optionally overrides the Host
	// header to send. If empty, the Request.Write method uses
	// the value of URL.Host.
	Host string		Host string
~~~

I'd still expect the Host being set in the executor Headers, adding a check for it and setting `req.Host` respectively.

titpetric:
Closes #217 

Signed-off-by: Tit Petric <tit.petric@monotek.net>

### DIFF
--- a/executors/http/executor.go
+++ b/executors/http/executor.go
@@ -102,6 +102,9 @@ func (Executor) Run(testCaseContext venom.TestCaseContext, l venom.Logger, step 
 
 	for k, v := range e.Headers {
 		req.Header.Set(k, v)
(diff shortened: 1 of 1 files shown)


titpetric:
Patch is tested locally, works as expected

yesnault:
LGTM
Thank you @titpetric

PR #222, opened 2020-02-13
feat(testfixtures): upgrade testfixtures to v3

wolf29f:
Hello,

There is a bug in testfixtures v2 with Postgres: if you have in your database a case sensitive index column (created with double quote) and want to use the dbfixtures executor to insert data you will get an error of the kind:

```
pq: relation "public.myindex does not exist
```
while the column is `myIndex`

The problem occurs even if the case sensitive index is on another table than the one you are inserting data into, because testfixtures reset indexes before inserting data.

Using the v3 of testfixtures fix the problem.


wolf29f:
Sorry, my mistake, it still doesn't work...

wolf29f:
I added a skipResetSequences parameter in dbfixtures to allow use of the executor despite the problem with case sensitivity

wolf29f:
Is it good for merge ?

PR #231, opened 2020-03-08
Feature - http -  Add a system to manage testSuite variable with body file

kevinramage:


yesnault:
Hi @kevinramage 

The goal with the PR is to "interpolate" the value of some variable, in the content of a file before POST it. Why not do the "interpolate" outside the executor, with the executor script for example ?

We are not very ready for this kind of modification given the refactoring of the executors and the contexts to come on the refactor branch.

kevinramage:
> Hi @kevinramage
> 
> The goal with the PR is to "interpolate" the value of some variable, in the content of a file before POST it. Why not do the "interpolate" outside the executor, with the executor script for example ?
> 
> We are not very ready for this kind of modification given the refactoring of the executors and the contexts to come on the refactor branch.

Hi,
We actually use a external python system to generate JSON files from a template to generate unique ID in our POST requests. I see this workaround not help developpers to understand the venom concept. That why I try to find a way to integrate this interpolation directly in Venom. I think, I will keep this not official way to do for the developpers and I will wait the venom refactoring to implement it properly after.
I close this PR, I will propose a new way to do it after the refactoring I think.
Thanks for you answer,
Regards

PR #233, opened 2020-03-16
Feature - Function - Add unique and random ID functions

kevinramage:


yesnault:
Thank you for this proposal.

On the refactor branch, it will be possible to do something like that:

```
{{ randNumeric 10 }}
```
The template engine will use the CDS template, so it will possible to use many helpers, see some doc on https://ovh.github.io/cds/docs/concepts/variables/#helpers

We prefer to not merged your PR on master, it will be in conflict with the incoming refactor.

kevinramage:
> Thank you for this proposal.
> 
> On the refactor branch, it will be possible to do something like that:
> 
> ```
> {{ randNumeric 10 }}
> ```
> 
> The template engine will use the CDS template, so it will possible to use many helpers, see some doc on https://ovh.github.io/cds/docs/concepts/variables/#helpers
> 
> We prefer to not merged your PR on master, it will be in conflict with the incoming refactor.

OK thanks for the answer, the main feature I need is the unique integer ID, I don't have the good solution I think for that. I will wait the helpers new feature to propose a new PR.
Regards

PR #235, opened 2020-03-17
fix(process): don't ignore invalid venom testsuites #234

### LINKED ISSUE #234: bug(processor): YAML syntax error ignored
Hi,

Hi, when I run venom against a list of testsuite files with one file not well formatted, the syntax error is ignored and venom returns a success.

I create two TestSuite files: 

MyBadTestSuite.yml
```
name: Title of TestSuite
testcases:
- name: TestCase with default value, exec cmd. Check if exit code != 1
  steps:
    - script: echo 'foo'
    type: exec
```

and MyGoodTestSuite.yml

```
name: Title of TestSuite
testcases:
- name: TestCase with default value, exec cmd. Check if exit code != 1
  steps:
  - script: echo 'foo'
    type: exec
```
if i run venom with a single file the behaviour is correct.

```bash
$ venom run MyBadTestSuite.yml
FATA[0000] yaml: line 5: did not find expected '-' indicator 
```

```
$ venom run MyGoodTestSuite.yml 
SUCCESS MyGoodTestSuite.yml                            1.264392ms

Total:1 Duration:1.823555ms
OK:1
KO:0
Skipped:0
TestSuite:1
TestCase:1
```
However, when I run with these two files no errors and no warning are yielding.

```
$ venom run MyBadTestSuite.yml MyGoodTestSuite.yml
SUCCESS MyGoodTestSuite.yml                            1.171967ms

Total:1 Duration:1.861674ms
OK:1
KO:0
Skipped:0
TestSuite:1
TestCase:1
```
I try with the branch master (c1d7719 -  feat(dbfixtures): add skipResetSequences parameter) without success.

youen:
close #234

### DIFF
--- a/cli/venom/run/cmd.go
+++ b/cli/venom/run/cmd.go
@@ -50,6 +50,7 @@ var (
 	parallel        int
 	stopOnFailure   bool
 	enableProfiling bool
+	noCheckYAML     bool
 	v               *venom.Venom
 )
 
@@ -66,6 +67,7 @@ func init() {
 	Cmd.PersistentFlags().StringVarP(&logLevel, "log", "", "warn", "Log Level : debug, info or warn")
 	Cmd.PersistentFlags().StringVarP(&outputDir, "output-dir", "", "", "Output Directory: create tests results file inside this directory")
 	Cmd.PersistentFlags().BoolVarP(&enableProfiling, "profiling", "", false, "Enable Mem / CPU Profile with pprof")
+	Cmd.PersistentFlags().BoolVarP(&noCheckYAML, "no-check-yml", "", false, "Don't check YML files before run and ignore invalid venom suite")
 }
 
 // Cmd run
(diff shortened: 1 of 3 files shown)


yesnault:
HI @youen 

Thank you for this pull request.

We want to keep the existing behaviour as the default behaviour, as some people use some fixtures in yml format, not valid with venom, in the same directory as the test file.

But, it should be great to add a flag as `--disabled-yml-check` (so with the default setted to false).

alexisvisco:
This PR should be reviewed because it's a **critical bug** which leads to not tested files.

yesnault:
Hi all. Yes, we are aware of this pull-request. It's under discussion and we will rework in few weeks. Our goal is to not impact existing users and if we can avoid adding a flag to enable or not a feature, it should be better.
With this PR implementation of the check, the check is done when the file is read. But, with some templating variables, the check will probably failed even if the file is ok with the good values of variables. It's not so easy to do a good check with different use case that we already know, that's why removing the feature was better than trying to get something to work, but that ruled out some uses of the template engine.

adrienaury (approved):

yesnault:
Ok. We discussed this issue again, we decided to rollback the current behavior which ignore errors in yml files. We don't want a flag to ignore or not: if some users want to run only a subset of yml files, they will have to filter the files list given to venom.
We will release the venom  0.28.0 with the current master, then removing the 'no-check' for the release 0.29.

Issue #242, opened 2020-03-30
feat(executor/http): return compact json body


reno-xjb:
If the http endpoint is not returning a compact json body, then using `result.body` in an exec step (to pipe it in `jq` for example) is causing an error due to the linebreaks breaking the YAML structure:
```
Error while unmarshal: yaml: line ##: did not find expected key
```

Proposal: compact the body and return it in `result.compactbody`, for backward compatibility
```go
var compactBody bytes.Buffer
err = json.Compact(&compactBody, bb)
if err == nil {
  r.CompactBody = compactBody.String()
}
```

```go
type Result struct {
	Executor    Executor    `json:"executor,omitempty" yaml:"executor,omitempty"`
	TimeSeconds float64     `json:"timeseconds,omitempty" yaml:"timeseconds,omitempty"`
	TimeHuman   string      `json:"timehuman,omitempty" yaml:"timehuman,omitempty"`
	StatusCode  int         `json:"statuscode,omitempty" yaml:"statuscode,omitempty"`
	Body        string      `json:"body,omitempty" yaml:"body,omitempty"`
	CompactBody string      `json:"compactbody,omitempty" yaml:"compactbody,omitempty"`
	BodyJSON    interface{} `json:"bodyjson,omitempty" yaml:"bodyjson,omitempty"`
	Headers     Headers     `json:"headers,omitempty" yaml:"headers,omitempty"`
	Err         string      `json:"err,omitempty" yaml:"err,omitempty"`
}
```

I could do a PR if you accept this change.

PR #244, opened 2020-03-31
feat(executor/http): add compactbody output with compacted JSON response

reno-xjb:
#242

### DIFF
--- a/executors/http/README.md
+++ b/executors/http/README.md
@@ -79,6 +79,7 @@ result.timeseconds
 result.timehuman
 result.statuscode
 result.body
+result.compactbody
 result.bodyjson
 result.headers
 result.error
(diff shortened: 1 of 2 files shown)


yesnault:
Hi @reno-xjb 
I'm not that we want to do that. This will create a new venom variable with all data duplicated from bodyJson. This can cause some perf issue if you manipulate a lot a variables with big json data. Venom had a lot of duplicated value, extracted from body / bodyJson, it's probaly better to work with them when you want without having to create compactJson every time.
If you want to compact json and use it as a new variable, you can create a new step script, to compact the json with a sh command, and export value in a new variable.

In the next major release, venom will use the interpolate from CDS. The compactJson can be added here : https://github.com/ovh/cds/blob/master/sdk/interpolate/interpolate_helper.go

This will let user to do something like that: `{{.aa.bodyjson|compactJson}}`, and this will be available on all executors playing with json, not only the http executor.

reno-xjb:
I understand your point and it was not my intention to bloat the codebase.
Nevertheless, as I stated in #242, I can't create a new step script to compact the body myself as it is breaking the YAML structure.
I'm not a big fan of API returning non-compacted JSON body, but I have to deal with some, so I will keep using my fork for now and I'm eager to use these interpolation commands.

Issue #243, opened 2020-03-31
feat: allow multiple variable files


reno-xjb:
It would be really useful to allow multiple variable files in order to separate variables or overwrite global variables (i.e. last file variables take precedence over previously defined variables):
```bash
venom run --var-from-file global_vars.yml --var-from-file my_vars.yml
```
Could it be considered?

nqb:
@reno-xjb, with current state of Venom there is two problems:

1. `--var-from-file` can be use only once on the command line. Consequently, you can have only one file to put all your variables.

2. override a variable on the command line with `--var` is not possible if you already use `--var-from-file` flag

With the title and description of your issue, I'm not sure which one you want to get fixed.

I need these two features !

reno-xjb:
@nqb I didn't know the second issue as I don't use `--var` myself, so I was more focused on the first one

nqb:
@reno-xjb, see my comments in #245

PR #247, opened 2020-04-01
feat(cli): allow multiple occurence of --var-from-file

reno-xjb:
#243

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .idea
 .vscode
 cli/venom/bin
+venom
 tests/pprof*
 tests/.DS_Store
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,6 +1,6 @@
 .idea
 .vscode
-venom
+cli/venom/bin
```
yesnault, line 3:
```suggestion
cli/venom/bin
venom
```

yesnault: ok to ignore bin, but keep ignore venom to avoid push the bin venom produced with a simple go build in cli/venom directory

PR #251, opened 2020-04-16
feat(http): read proxy from environment

moepi:
I've added support for reading proxy configuration from environment variables.


moepi:
I see there are some open pull requests created a while ago. When do you think will this be merged?

yesnault:
@moepi yes, sorry for delay. We should work on this next week,

PR #273, opened 2020-10-07
fix: do not truncate filename in results

### LINKED ISSUE #254: bug: venom doesn't display full path name of test suite
If you use an absolute path to a file which is long, time execution overrides path name.

### Steps to reproduce

1. Create a test suite with a loooong absolute path
```console
$ mkdir -p /tmp/path/to/a/long/filename/
$ cat >> /tmp/path/to/a/long/filename/test_suite_name.yml << EOF
> name: Title of TestSuite
> testcases:
> - name: Check if exit code != 1 and echo command response in less than 1s
>   steps:
>   - script: echo 'foo'
>     assertions:
>     - result.code ShouldEqual 0
>     - result.timeseconds ShouldBeLessThan 1
> EOF
```
2. Run tests
```console
$ venom run /tmp/path/to/a/long/filename/test_suite_name.yml
SUCCESS /tmp/path/to/a/long/filename/test_suite_name.ym13.081607ms
[..]
```

### Actual results
We can see that last character of filename (`l`) is not displayed

### Expected results
Display full path name.

yesnault:
close #254

### DIFF
--- a/process_testsuite.go
+++ b/process_testsuite.go
@@ -70,12 +70,12 @@ func (v *Venom) runTestSuite(ts *TestSuite) {
 	var o string
 	if ts.Failures > 0 || ts.Errors > 0 {
 		red := color.New(color.FgRed).SprintFunc()
-		o = fmt.Sprintf("%s %s", red("FAILURE"), rightPad(ts.Package, " ", 47))
+		o = fmt.Sprintf("%s %s", red("FAILURE"), ts.Package)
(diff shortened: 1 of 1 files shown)


yesnault:
before:
Capture d’écran 2020-10-07 à 13 52 48

after:

Capture d’écran 2020-10-07 à 13 54 22

fsamin (approved):

PR #278, opened 2020-10-09
Fix issue 267 - Add wrapper assertion functions to parse string to int

### LINKED ISSUE #267: Assertion ShouldBeLessThan fails
Hello,

It seems that ShouldBeLessThan assertion is not working.

I'm using venom v0.27.0 and the following example:
`version: "2"

name: Test Assertions

testcases:
- name: TestAssertions
  steps:
  - script: echo 3
    assertions:
    - result.systemout ShouldBeGreaterThan 1
    - result.systemout ShouldBeLessThan 10`

I get the error: "Expected '3' to be less than '10'".
I'm not very good in maths, but I think there is a problem here :-)

I tried to use the function go assertions.ShouldBeLessThan and it's working as expected so I think it comes from the venom integration.

Thanks for this wonderful tool.
Luis

kevinramage:
close #267


yesnault:
@kevinramage Thank you, just two things:
 - can you add some unit tests in assertion_test.go file
 - can you sign your commit https://github.com/ovh/venom/blob/master/CONTRIBUTING.md#submitting-modifications

yesnault:
fixed by https://github.com/ovh/venom/pull/286

PR #279, opened 2020-10-09
Change the way to decode json to change the way to parse big number

### LINKED ISSUE #270: Bug: wrong formating for number
Hello,

When using a number returned from a previous step (eg: an id returned by an element creation), if the id is too long, when using it in another step it is rendered with scientific writing, which can cause trouble when testing (eg: updating a previously created element).

This behavior is pretty easy to reproduce, create a simple client, for example here's mine:

```go
func main() {

	server := http.Server{}
	server.Handler = http.HandlerFunc(func(resp http.ResponseWriter, req *http.Request) {
		resp.Header().Add("content-type", "application/json")
		data := map[string]interface{}{
			"id": 1167860,
		}
		respBody, err := json.Marshal(data)
		if err != nil {
			log.Error(err)
		}

		_, err = resp.Write(respBody)
		if err != nil {
			log.Error(err)
		}
	})
	server.Addr = "127.0.0.1:8080"
	server.ListenAndServe()
}
```

And a simple test file

```yml
name: "Test venom"
testcases:
  - name: "httpGet"
    steps:
      - type: http
        method: GET
        url: "http://127.0.0.1:8080"
        assertions:
          - "result.bodyjson.id ShouldEqual 1167860"
  - name: "print"
    steps:
      - type: exec
        script: >
          echo "Body content: '{{.httpGet.result.bodyjson.id}}'"
```

When running with debug log level we can see that the assertion for the http step work as expected, but the exec will print out `Body content: '1.16786e+06'` instead of `Body content: '1167860'`, the same happen when using the value in a URL.

kevinramage:
To fix the issue, you must change the way to decode JSON content, I use a specific decoder with the property useNumber set to true.

close #270

### DIFF
--- a/executors/http/executor.go
+++ b/executors/http/executor.go
@@ -157,14 +157,11 @@ func (Executor) Run(testCaseContext venom.TestCaseContext, l venom.Logger, step 
 			}
 			r.Body = string(bb)
 
-			bodyJSONArray := []interface{}{}
-			if err := json.Unmarshal(bb, &bodyJSONArray); err != nil {
-				bodyJSONMap := map[string]interface{}{}
-				if err2 := json.Unmarshal(bb, &bodyJSONMap); err2 == nil {
-					r.BodyJSON = bodyJSONMap
-				}
-			} else {
-				r.BodyJSON = bodyJSONArray
+			var m interface{}
+			decoder := json.NewDecoder(strings.NewReader(string(bb)))
+			decoder.UseNumber()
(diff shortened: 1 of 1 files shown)


yesnault:
@kevinramage Thank you. LGTM.
Can you sign your commit can you sign your commit https://github.com/ovh/venom/blob/master/CONTRIBUTING.md#submitting-modifications please?

yesnault:
close https://github.com/ovh/venom/issues/270

PR #183, opened 2019-03-21
refactor all venom

fsamin:
- [ ] Add goals in makefile to perform integration tests
- [ ] Refactor matching os/architecture in Makefile
- [ ] Implement context module SSH - which is able to handle itself some executor type such as `exec`
- [ ] Implement context module Redis - which is able to handle itself some executor type `redis`
- [ ] Variables parsing before run
- [ ] Test curse output on other architecture/terminals
- [ ] Support for windows architecture
- [ ] Development docs
- [ ] Output modules (to be able to implement nice html report module)
- [x] Backport https://github.com/ovh/venom/commit/6aebf36775b6041c94c41a0d75442977405f85c7
- [x] Backport https://github.com/ovh/venom/commit/0b9f7ea1fe0f6e008f0ed12d59b08ec935f186ec
- [x] Backport https://github.com/ovh/venom/commit/435591135bc0ae903bc8d728b7777f403b452ce3
- [ ] Backport 62a00a9b9e013bc5da81bd6bbecd49e746c737dc
- [ ] Backport 0d2ed76f4a0f44782a32dbb39beac54c8b4ed86b
- [ ] Backport 24e4e6c4a03046e4918bd4b80e5bd05f8b9bf9e8
- [ ] Backport 5845416c51b2cdb6f18e1c07bd0ee2069c2c35e8
- [ ] Backport 749a51cdea6ff6252464bcd2a34b1532f68f288d
- [ ] Backport dc7ae075fb691dcb2713b7342369f3e1f9fe7992
- [ ] Backport f9547b995828103d9a38678d7393a8deb359c254

PR #289, opened 2020-10-26
Feat/makefile

fabriziomoscon:
The contribution instruction suggest to specific commands like `vet` and `lint`. Where go vet is standard in go across environment. Linting must be declared in a deterministic way, to ensure that all developers and CI are fixing the same rules.

`make deps` ensures that `go mod tidy` is always run when installing dependencies and it tidy up `go.mod` and `go.sum`, this prevents future contributors to find unrelated changes in their PRs because they base commit had untidy dependencies.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,76 @@
+# options for analysis running
+run:
+  # default concurrency is a available CPU number
+  concurrency: 4
+
+  # timeout for analysis, e.g. 30s, 5m, default is 1m
+  deadline: 2m
+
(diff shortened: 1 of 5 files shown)


fabriziomoscon:
@fsamin 
I only saw now PR #183 
Perhaps this could be a stepping stone to start using a simple Makefile. Anyhow, the target lint is not inside the refactor, so there might be value on a couple of these commits.

fsamin (approved):
thanks for your PR
lgtm

yesnault (approved):

PR #300, opened 2020-11-16
fix(executor/sql): removing odbc for now

yesnault:
until executor as plugin is available, this will let everybody compile
and run venom without installing unixodbc library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,9 +9,6 @@ It can also output xUnit results files.
 
 Download latest binary release from https://github.com/ovh/venom/releases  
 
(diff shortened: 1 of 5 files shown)


ovh-cds:
CDS Report build-venom#281.0 ✘  
* Build
  * Build ✔ 
* Tests
  * Acceptance Tests ✘  
  * Lib Unit Tests ✔

yesnault:
@kevinramage FYI, we had to remove odbc for now, to let users run venom without having unixodbc installed. We'll add support to go plugin soon (should work on it this week), this will let venom to have a executor odbc as `.so` file, that users can load only if they need it.

fsamin (approved):

PR #398, opened 2021-04-16
💡  Proposal: Add testsuite labels

manzan46:
⚠️  Feel free to discard the PR if you think it should not be included. 

💡  So the idea is that we want to easily differentiate test suites, some could run in CI while some others could be run manually, some are critical while others are not. 

**Proposal: we would like to be able to via a command or env var to be able to specify these `labels`** 

This way we can specify which test suite can be run or not, if no labels pass as option everything is run :) 

so with the following testsuites: 

```
# testsuite1.yml
name: a_critical_scenario
labels:
 - critical
vars: 
 ...
testcases:
- name: test_case1
- name: test_case2

# testsuite2.yml
name: a_non_critical_scenario_with_label
labels:
 - legacy
vars: 
 ...
testcases:
- name: test_case1
- name: test_case2

# testsuite3.yml
name: a_non_critical_scenario
vars: 
 ...
testcases:
- name: test_case1
- name: test_case2
```

- `venom run *.yml` would run them all
- `venom run *.yml --label critical` would only run the first one
- `venom run *.yml --label critical --label legacy` would only run the first two 

---

ℹ️  If you're happy with adding this feature, I can finish up PR
❓ One thing I'm unsure is how to handle the `computeStats` part as there is already a `Skipped` for `Testsuite` but that seems to be incremented when testcase steps are skipped.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,8 @@ Flags:
       --format string           --format:yaml, json, xml, tap (default "xml")
   -h, --help                    help for run
       --lib-dir string          Lib Directory: this directory can contain user executors. This overrides the default lib folder directory
+      --label strings           --label critical --label ci-only: only
+      run testsuites with specified labels.
       --output-dir string       Output Directory: create tests results file inside this directory
       --stop-on-failure         Stop running Test Suite on first Test Case failure
       --var stringArray         --var cds='cds -f config.json' --var cds2='cds -f config.json'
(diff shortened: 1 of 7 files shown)


yesnault:
Thank you for this PR.

We prefer to not add a flag for this feature on venom. This can be achieved by many ways as a good local organization of files. 

Example with symlinks:

```
tree
.
├── label_1
│   ├── test_a.yml -> ../test_a.yml
│   ├── test_b.yml -> ../test_b.yml
│   └── test_c.yml -> ../test_c.yml
├── label_2
│   ├── test_d.yml -> ../test_d.yml
│   └── test_e.yml -> ../test_e.yml
├── test_a.yml
├── test_b.yml
├── test_c.yml
├── test_d.yml
└── test_e.yml

```

Other solution:

```
❯ tree
.
├── test_critical_1.yml
├── test_legacy_1.yml
├── test_legacy_2.yml
└── test_legacy_3.yml

0 directories, 4 files
❯ venom run test_legacy_*.yml

manzan46:
Ok make sense thank you for your reply 🙏🏻

PR #404, opened 2021-06-02
Parse Redis responses containing nested arrays

chris1786:
The Redis executor makes the assumption that if Redis returns an array as a response, this array will be flat, and there will be no arrays nested within it. This is not always the case (for example, **XREAD** and **XRANGE** return nested arrays).

This PR amends the Redis executor to parse Redis responses containing nested arrays, and includes a test-case to demonstrate how this functionality can be used with Redis steams.

Signed-off-by: chris1786 <22242927+chris1786@users.noreply.github.com>


fsamin (approved):

yesnault:
@chris1786  Thank you!

yesnault (approved):

PR #414, opened 2021-07-03
Fix example executor code

maxstepanov:
Found some errors in executor example while exploring


yesnault:
@maxstepanov  thank you!

yesnault (approved):

PR #431, opened 2021-10-05
Added Tavern executor

c4s4:
Added new [Taven](https://tavern.readthedocs.io/en/latest/) like executor. See documentation at <https://github.com/intercloud/venom/tree/executor-tavern/executors/tavern>.


yesnault:
Hi @c4s4
After consideration (really sorry for the response time), this system is really interesting. But, we are not fan of creating a new executor to introduce a new assertion system, only for this executor (mention that the http executor is also evolving on its side). I rather prefer to evolve the core of venom to set it up, with a new attribute `response` in assertions field for example.

PR #448, opened 2021-11-11
Improve Kafka executor

philippgille:
Hello 👋 ,

I think I found one potential issue and one improvement for the Kafka executor:

I checked with a topic that has multiple partitions, and Sarama starts multiple goroutines in which the `handler` is used, so `ConsumeClaim` is called from and running in multiple goroutines concurrently. With that:

1. It can happen that despite a `messageLimit` of 1, after one goroutine added a message to `h.messages` and unlocks the mutex, another goroutine could enter and add another message.
    - => Checking the message length _before_ adding another one should fix this
2. Let's say we use a key for filtering, and one goroutine found a matching msg and hits the `messageLimit` and returns. But other goroutines still continue running, because with the `continue` in the key filter condition it never reaches the limit check. I'm not sure if Sarama notices that one goroutine returned and eventually closes the `claim.Messages()` channel itself, because at least the other goroutines don't seem to continue their full `timeout`/`waitTime` duration, but when adding some logs I definitely saw them continue to consume some msgs after one goroutine hit the `messageLimit`.
    - => Using a channel for signaling that the limit has been met should fix this
 
I can split this into two PRs if you prefer that.

### DIFF
--- a/executors/kafka/kafka.go
+++ b/executors/kafka/kafka.go
@@ -315,6 +315,7 @@ func (e Executor) consumeMessages(ctx context.Context) ([]Message, []interface{}
 		messageLimit: e.MessageLimit,
 		schemaReg:    e.schemaReg,
 		keyFilter:    e.KeyFilter,
+		done:         make(chan struct{}),
 	}
 	if err := consumerGroup.Consume(ctx, e.Topics, h); err != nil {
 		if e.WaitFor > 0 && errors.Is(err, context.DeadlineExceeded) {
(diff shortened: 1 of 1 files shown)


sixstone-qq (approved):
Nice fix . It looks great to me :hugs:

yesnault (approved):
LGTM

fsamin (approved):

ovh-cds:
CDS Report build-venom-a#677.0 ✘  
* Build
  * Build ✔ 
* Tests
  * Acceptance Tests ✘

philippgille:
> CDS Report build-venom-a#677.0 ✘
> Acceptance Tests ✘

Where can I see the CDS logs?

yesnault:
> > CDS Report build-venom-a#677.0 ✘
> > Acceptance Tests ✘
> 
> Where can I see the CDS logs?

Hi, 

It's ok now
" build-venom-a: Success"

Thank you for this PR.

PR #453, opened 2021-11-25
feat(tests): add `range` supports

### LINKED ISSUE #263: feat: Running same test case for list of inputs
It would be nice to be able to set a list of inputs that would be run for the same testcase instead of having to repeat the same testcase N times.

Example:

```yaml
testcases:
  - name: testA
    inputs:
      - inputA: "1"
        inputB: "abc"
      - inputA: "2"
        inputB: "def"
    type: exec
    script: echo "first input is {{.inputA}}, second input is {{.inputB}}
```

### LINKED ISSUE #288: Possibility to iterate over an array in a step
It could be great to iterate over an array (declared in a variable or a result of a previous step) inside a step to test some different datasets.
Example : You have a first step that return an array

```YAML
name: test

testcases:

- name: getArray
  steps:
  - type: http
    method: GET
  ...
```

In the next step you want to iterate over the results 
```YAML
name: test

testcases:

- name: testOnAllArray
  steps:
  - loopover : result.getArray.bodyjson
  - loopvar : myData #take the value of the array element
  - type: http
    method: GET
    url : https://getmyid?id:eq={{.myData}}
  ...
```

lowlighter:
May close #263 and close #288

This add support for the `range` attribute that can be passed in a step context.
The following types along their contexted variables are supported:
- `[]interface{}`: an array of values
  - `.index`/`.key`: current iteration index
  - `.value`: current iteration item value
- `map[string]interface{}`: a map of values (each key will be iterated over)
  - `.index`: current iteration index
  - `.key`: current iteration item key
  - `.value`: current iteration item value
- `int`: an integer to perform action `n` times
  - `.index`/`.key`/`.value`: current iteration index
- `string`: a templated string which results in one of the above typing 
  - It can be either inherited from vars file, or interpolated from a previous step result 

Within a range context, `.range` variable will be set to `true` (while it'll be `false` in "regular" contexts)
 
Here's an example (see `tests/ranged.yml` in this PR diff for more examples):
```yaml
- name: range with harcoded array
  steps:
  - type: exec
    range:
      - actual: hello
        expected: hello
        index: 0
      - actual: world
        expected: world
        index: 1
    script: echo "{{.range}} {{.index}} {{.value.actual}}"
    assertions:
    - result.code ShouldEqual 0
    - result.systemout ShouldEqual "true {{.value.index}} {{.value.expected}}"
```
Image below show the content of `Range struct` for each test case:
image

____

I haven't run `go fmt` yet as it'll enclose the current step processing (L176 to L281) within the `for` which make the diff harder to review. If no further changes are requested, I'll run the formatter and add some documentation in the `README.md` with definitive syntax :slightly_smiling_face:

### DIFF
--- a/README.md
+++ b/README.md
@@ -527,8 +527,6 @@ The following data types are supported, each exposing contexted variables `.inde
 - A templated string which results in one of the above typing (`string`)
   - It can be either inherited from vars file, or interpolated from a previous step result
 
-Within a range context, `.range` contexted variable will be set to `true` (while it'll be `false` in regular contexts)
-
 For instance, the following example will iterate over an array of two items containing maps:
 ```yaml
 - name: range with harcoded array
@@ -537,14 +535,12 @@ For instance, the following example will iterate over an array of two items cont
     range:
       - actual: hello
         expected: hello
-        index: 0
       - actual: world
         expected: world
-        index: 1
-    script: echo "{{.range}} {{.index}} {{.value.actual}}"
+    script: echo "{{.value.actual}}"
     assertions:
     - result.code ShouldEqual 0
-    - result.systemout ShouldEqual "true {{.value.index}} {{.value.expected}}"
+    - result.systemout ShouldEqual "{{.value.expected}}"
 ```
 
 More examples are available in [`tests/ranged.yml`](/tests/ranged.yml).
--- a/README.md
+++ b/README.md
@@ -24,6 +24,7 @@ It can also generate xUnit result files.
 * [Advanced usage](#advanced-usage)
   * [Debug your testsuites](#debug-your-testsuites)
   * [Skip testcase](#skip-testcase)
+  * [Iterating over data](#iterating-over-data)
 * [Use venom in CI](#use-venom-in-ci)
 * [Hacking](#hacking)
 * [License](#license)
@@ -507,6 +508,46 @@ testcases:
     - result.code ShouldEqual 0
 
 ```
+
+## Iterating over data
+
+It is possible to iterate over data using `range` attribute.
+
+The following data types are supported, each exposing contexted variables `.index`, `.key` and `.value`:
+
+- An array where each value will be iterated over (`[]interface{}`)
+  - `.index`/`.key`: current iteration index
+  - `.value`: current iteration item value
+- A map where each key will be iterated over (`map[string]interface{}`)
+  - `.index`: current iteration index
+  - `.key`: current iteration item key
+  - `.value`: current iteration item value
+- An integer to perform target step `n` times (`int`)
+  - `.index`/`.key`/`.value`: current iteration index
+- A templated string which results in one of the above typing (`string`)
+  - It can be either inherited from vars file, or interpolated from a previous step result
+
+Within a range context, `.range` contexted variable will be set to `true` (while it'll be `false` in regular contexts)
+
+For instance, the following example will iterate over an array of two items containing maps:
+```yaml
+- name: range with harcoded array
+  steps:
+  - type: exec
+    range:
+      - actual: hello
+        expected: hello
+        index: 0
+      - actual: world
+        expected: world
+        index: 1
+    script: echo "{{.range}} {{.index}} {{.value.actual}}"
+    assertions:
+    - result.code ShouldEqual 0
+    - result.systemout ShouldEqual "true {{.value.index}} {{.value.expected}}"
+```
+
+More examples are available in [`tests/ranged.yml`](/tests/ranged.yml).
 
 # FAQ
 
--- a/process_testcase.go
+++ b/process_testcase.go
@@ -181,113 +181,113 @@ func (v *Venom) runTestSteps(ctx context.Context, tc *TestCase) {
 				stepVars.Add("value", rangedData.Value)
 			}
 
-		vars, err := DumpStringPreserveCase(stepVars)
-		if err != nil {
-			Error(ctx, "unable to dump testcase vars: %v", err)
-			tc.AppendError(err)
-			return
-		}
-
-		for k, v := range vars {
-			content, err := interpolate.Do(v, vars)
+			vars, err := DumpStringPreserveCase(stepVars)
 			if err != nil {
+				Error(ctx, "unable to dump testcase vars: %v", err)
 				tc.AppendError(err)
-				Error(ctx, "unable to interpolate variable %q: %v", v, err)
 				return
 			}
-			vars[k] = content
-		}
 
-		// the value of each var can contains a double-quote -> "
-		// if the value is not escaped, it will be used as is, and the json sent to unmarshall will be incorrect.
-		// This also avoids injections into the json structure of a step
-		for i := range vars {
-			vars[i] = strings.ReplaceAll(vars[i], "\"", "\\\"")
-		}
-
-		var content string
-		for i := 0; i < 10; i++ {
-			content, err = interpolate.Do(string(rawStep), vars)
-			if err != nil {
-				tc.AppendError(err)
-				Error(ctx, "unable to interpolate step: %v", err)
-				return
-			}
-			if !strings.Contains(content, "{{") {
-				break
-			}
-		}
-
-		Info(ctx, "Step #%d content is: %q", stepNumber, content)
-		var step TestStep
-		if err := yaml.Unmarshal([]byte(content), &step); err != nil {
-			tc.AppendError(err)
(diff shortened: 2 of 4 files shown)


process_testcase.go
```diff
@@ -270,7 +287,100 @@ func (v *Venom) runTestSteps(ctx context.Context, tc *TestCase) {
...
+		}
+	}
+}
+
+//Parse and format range data to allow iterations over user data
+func parseRanged(ctx context.Context, rawStep []byte, stepVars H) (Range, error) {
+
+	//Load "range" attribute and perform actions depending on its typing
+	var ranged Range
+	json.Unmarshal([]byte(rawStep), &ranged)
```
fsamin, line 299:
```suggestion
	if err := json.Unmarshal(rawStep, &ranged); err != nil {
	     return nil, fmt.Errorf("unable to parse range expression: %v", err)
	} 
```

process_testcase.go
```diff
@@ -270,7 +287,100 @@ func (v *Venom) runTestSteps(ctx context.Context, tc *TestCase) {
...
+		// Try parsing already templated data
+		err := json.Unmarshal([]byte("{\"range\":"+rawString+"}"), &ranged)
+
+		// ... or fallback
+		if err != nil {
+			//Try templating and escaping data
+			Debug(ctx, "attempting to template \"range\" attribute and parse it again")
+			vars, err := DumpStringPreserveCase(stepVars)
+			if err != nil {
+				break
```
fsamin, line 325: you should log the error

process_testcase.go
```diff
@@ -270,7 +287,100 @@ func (v *Venom) runTestSteps(ctx context.Context, tc *TestCase) {
...
+			}
+			for i := range vars {
+				vars[i] = strings.ReplaceAll(vars[i], "\"", "\\\"")
+			}
+			content, err := interpolate.Do(string(rawStep), vars)
+			if err != nil {
+				break
```
fsamin, line 332: you should log the error

process_testcase.go
```diff
@@ -270,7 +287,100 @@ func (v *Venom) runTestSteps(ctx context.Context, tc *TestCase) {
...
+			}
+
+			//Try parsing data
+			err = json.Unmarshal([]byte(content), &ranged)
+			if err != nil {
+				break
```
fsamin, line 338: you should log the error

process_testcase.go
```diff
@@ -270,7 +287,100 @@ func (v *Venom) runTestSteps(ctx context.Context, tc *TestCase) {
...
+			}
+			switch ranged.RawContent.(type) {
+			case string:
+				rawString = ranged.RawContent.(string)
+				json.Unmarshal([]byte("{\"range\":"+rawString+"}"), &ranged)
```
fsamin, line 343: you should check the error

process_testcase.go
```diff
@@ -270,7 +287,100 @@ func (v *Venom) runTestSteps(ctx context.Context, tc *TestCase) {
...
+		rawString := ranged.RawContent.(string)
+		if len(rawString) == 0 {
+			return ranged, fmt.Errorf("\"range\" has been specified without any data")
+		}
+
+		// Try parsing already templated data
+		err := json.Unmarshal([]byte("{\"range\":"+rawString+"}"), &ranged)
+
+		// ... or fallback
+		if err != nil {
```
fsamin, line 320:
```suggestion
		err := json.Unmarshal([]byte("{\"range\":"+rawString+"}"), &ranged)
		// ... or fallback
		if err != nil {
```

process_testcase.go
```diff
@@ -270,7 +287,100 @@ func (v *Venom) runTestSteps(ctx context.Context, tc *TestCase) {
...
+	case float64:
+		Debug(ctx, "\"range\" data is number-like")
+		upperBound := int(ranged.RawContent.(float64))
+		for i := 0; i < upperBound; i++ {
+			key := strconv.Itoa(i)
+			ranged.Items = append(ranged.Items, RangeData{key, i})
+		}
+
+	//Map-like data
+	case map[string]interface{}:
```
fsamin, line 369:
```suggestion
	case map[string]interface{},map[interface{}]interface{}:
```

process_testcase.go
```diff
@@ -270,7 +287,100 @@ func (v *Venom) runTestSteps(ctx context.Context, tc *TestCase) {
...
+		Debug(ctx, "\"range\" data is map-like")
+		for key, value := range ranged.RawContent.(map[string]interface{}) {
+			ranged.Items = append(ranged.Items, RangeData{key, value})
+		}
+
+	//Unsupported data format
+	default:
+		fmt.Println(ranged.RawContent)
```
fsamin, line 377: why fmt.Println ?

fsamin (changes requested):
Thanks for the PR.
Here are some feedback.

Can you also add a new section in the README ?

process_testcase.go
lowlighter: Is the aim of using `map[interface{}]interface{}` for complex mapping ?

Like: 
```yaml
range:
      ? {key: foo}
      : {actual: hello, expected: hello}
      ? {key: bar}
      : {actual: world, expected: world}	
```
If that's the case I don't think it's supported from what I've seen [go.dev/play/p/4yC3E54dLGy](https://go.dev/play/p/4yC3E54dLGy) though I may have missed something

process_testcase.go
lowlighter: was a leftover from debugging, good catch

lowlighter:
- Applied feedback changes and rebased to fix conflict with #452
- Updated the `README.md` to add section about `range` expression
- Also applied `go fmt` (which makes the diff harder to read because it changes the indents, can revert it temporarly for review if you want)

yesnault (approved):

process_testcase.go
fsamin: tbh i don't remember when/how it's possible to get `map[interface{}]interface{}` out of an unmarshaler, you can skip this comment, if you prefer

README.md
```diff
@@ -508,6 +509,46 @@ testcases:
...
+
+For instance, the following example will iterate over an array of two items containing maps:
+```yaml
+- name: range with harcoded array
+  steps:
+  - type: exec
+    range:
+      - actual: hello
+        expected: hello
+        index: 0
```
fsamin, line 540: avoid this field in the map example, it can be confusing to have `value.index` and `.index`

README.md
```diff
@@ -508,6 +509,46 @@ testcases:
...
+- A map where each key will be iterated over (`map[string]interface{}`)
+  - `.index`: current iteration index
+  - `.key`: current iteration item key
+  - `.value`: current iteration item value
+- An integer to perform target step `n` times (`int`)
+  - `.index`/`.key`/`.value`: current iteration index
+- A templated string which results in one of the above typing (`string`)
+  - It can be either inherited from vars file, or interpolated from a previous step result
+
+Within a range context, `.range` contexted variable will be set to `true` (while it'll be `false` in regular contexts)
```
fsamin, line 530: in fact i'm not sure the `{{.range}}` is still useful ? how do you think you will use it ?

fsamin (changes requested):

README.md
lowlighter: it's mostly for testing that `.index` is effectively templated and set to correct value, but from an user pov it makes no sense
maybe just renaming `index` to `testIndex` would be better ?

README.md
lowlighter: same as above, it's mostly for unit test to check in which context the step is running

README.md
fsamin: it's fine to keep it in our tests files. but avoid it in the README.md

README.md
fsamin: i guess the other tests can assert the whole feature is working. i prefer to remove it if it's not strictly required.

process_testcase.go
lowlighter: ok I'll let as it since I don't have a test case for it 
I don't think it's possible to make dynamic type assertions, so it would actually require to create a new case to handle specifically `ranged.RawContent.(map[interface{}]interface{})`

README.md
lowlighter: ok I removed it from `README.md`

README.md
lowlighter: ok I removed all references to `{{.range}}` and also removed the go code that was setting it

fsamin (approved):

PR #455, opened 2021-11-29
feat(assertions): add logical operators

### LINKED ISSUE #312: operator or in the definition of assertions

lowlighter:
Closes #312 

This add support for logical operators syntax.
Supported operators are `and`, `or` and `xor`.
Nested operators are also supported.

There's a `not` operator too but it was added mostly to test that operators work correctly in tests case. User should use the respective "Not" assertions functions as a best practice rather than use it, that's why I didn't really document this one.

```yml
- name: Assertions operators
  steps:
  - script: echo 1
    assertions:
      - or:
        - result.systemoutjson ShouldEqual 1 
        - result.systemoutjson ShouldEqual 2
      # Nested operators
      - or:
        - result.systemoutjson ShouldBeGreaterThanOrEqualTo 1
        - result.systemoutjson ShouldBeLessThanOrEqualTo 1
        - or:
          - result.systemoutjson ShouldEqual 1
```

image

### DIFF
--- a/tests/assertions_operators.yml
+++ b/tests/assertions_operators.yml
@@ -36,6 +36,17 @@ testcases:
           - or:
             - result.systemoutjson ShouldEqual 1
 
+  - name: AssertionsOperatorNested2
+    steps:
+    - script: echo 1
+      assertions:
+        - or:
+          - or:
+            - or:
+              - or:
+                - or:
+                  - result.systemoutjson ShouldEqual 1
+
   - name: AssertionsReadmeExample
     steps:
     - script: echo 1
--- a/tests/assertions_operators.yml
+++ b/tests/assertions_operators.yml
@@ -17,14 +17,6 @@ testcases:
           - result.systemoutjson ShouldEqual 1 
           - result.systemoutjson ShouldEqual 2
 
-  - name: AssertionsOperatorOrError
-    steps:
-    - script: echo 1
-      assertions:
-        - or:
-          - result.systemoutjson ShouldEqual 0 
-          - result.systemoutjson ShouldEqual 2
-
   - name: AssertionsOperatorXor
     steps:
     - script: echo 1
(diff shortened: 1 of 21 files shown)


lowlighter:
Few additional notes:
- `and` was added but is probably redundant as it's already implicit by design, maybe we shouldn't add it
- `not` was not added because it's better to force the use of "Not" assertions functions I think

The following diff introduces a breaking change for custom go-executor:
```diff
+ type Assertion interface{}
+ Assertions []Assertion `json:"assertions,omitempty" yaml:"assertions,omitempty"` 
- Assertions []string `json:"assertions,omitempty" yaml:"assertions,omitempty"`
```
It is however needed to allow users pass dict mapping in assertions expression in their YAML

tests/assertions_operators.yml
```diff
@@ -0,0 +1,68 @@
...
+          - result.systemoutjson ShouldEqual 1 
+          - result.systemoutjson ShouldEqual 2
+
+  - name: AssertionsOperatorOrError
+    steps:
+    - script: echo 1
+      assertions:
+        - or:
+          - result.systemoutjson ShouldEqual 0 
+          - result.systemoutjson ShouldEqual 2
```
lowlighter, line 26: is there a way to "expect assertions errors" ?
it's kind of meta but the aim here would be to test expected failing cases

ovh-cds:
CDS Report build-venom-a#683.0 ✘  
* Build
  * Build ✔ 
* Tests
  * Acceptance Tests ✘

tests/assertions_operators.yml
lowlighter: solved this by implementing the `not` operator

ovh-cds:
CDS Report build-venom-a#685.0 ✘  
* Build
  * Build ✔ 
* Tests
  * Acceptance Tests ✔ 
* Analyse
  * Sonar ✘

ovh-cds:
CDS Report build-venom-a#687.0 ✘  
* Build
  * Build ✔ 
* Tests
  * Acceptance Tests ✔ 
* Analyse
  * Sonar ✘

fsamin (approved):

fsamin:
i think we should release final 1.0.0 and merge this in 1.1.0 branch

yesnault (approved):

PR #456, opened 2021-11-30
Upgrade Kakfa client version to 2.6.0

sarunask:
Upgrade Kakfa client version to 2.6.0 as older versions do not work correctly with RedPanda, Kafka cluster alternative, written in GO.
We use RedPanda as an alternative to Confluent Kafka, due to fact it takes ~300Mb to download and run single fast GO service, compared to 2.5Gb of download and run 3 slow starting Java services, which are required to run Kafka.

### DIFF
--- a/executors/kafka/kafka.go
+++ b/executors/kafka/kafka.go
@@ -336,7 +336,7 @@ func (e Executor) getKafkaConfig() (*sarama.Config, error) {
 	config.Net.SASL.Password = e.Password
 	config.Consumer.Return.Errors = true
 	config.Net.DialTimeout = defaultDialTimeout
-	config.Version = sarama.V0_10_2_0
+	config.Version = sarama.V2_6_0_0
 
 	if e.KafkaVersion != "" {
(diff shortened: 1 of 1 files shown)


yesnault:
Hi @sarunask 
The min version compatible is V0_10_2_0. You can set your version manually in your testsuite. 
Force the default value to another value is a breaking change for all existing users. Why not let user set manually the version to 2.6 ?

sarunask:
> Hi @sarunask The min version compatible is V0_10_2_0. You can set your version manually in your testsuite. Force the default value to another value is a breaking change for all existing users. Why not let user set manually the version to 2.6 ?

I suppose no-one is using v0.10 of Kafka cluster anymore for few years, so it's quite obsolete version. It has bugs and security issues, which are not fixed.
V2.6 can be safe default for upcoming years, as all new Kafka clusters are at least this version or even newer, as now v3.1 is already available.

yesnault (approved):

marcaudefroy (approved):

Issue #465, opened 2021-12-13
Default xml output has no file extension


clement-sciascia:
When running venom without a 'format' option, the default output is XML.
However, the output file has no extension, it ends up being "test_results."

PR #466, opened 2021-12-13
Ensure the output file has '.xml' extension

clement-sciascia:
fix: when output defaults to XML, the file should have the '.xml' extension (#465)

Signed-off-by: Clement Sciascia <clement.sciascia@ovhcloud.com>

I wasn't sure where to put the default value, feel free to suggest a better way or edit my commit.
The default 'xml' value is already defined in the function that instantiates a Venom struct, but gets overwritten by this 'format' variable after flags parsing

### DIFF
--- a/cmd/venom/run/cmd.go
+++ b/cmd/venom/run/cmd.go
@@ -58,6 +58,8 @@ var (
 )
 
 func init() {
+	format = "xml"
+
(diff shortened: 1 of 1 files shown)


fsamin (approved):
Thanks for you PR

PR #478, opened 2021-12-27
fix: interpolation lib bad µµµ replacement

### LINKED ISSUE #475: Test case interpolation incorrectly replaces "-" with "µµµ"
String interpolation seems to break in a very specific edge-case. 

I've been able to create a minimal test suite to reproduce the problem in Venom 1.0.0, 1.0.1, and the latest from `master` (1.0.0-rc.7 and older don't have this issue):

```yaml
---
name: Interpolation POC
vars:
  accept: "text/html"

testcases:
  - name: GetVar
    steps:
      - type: http
        method: GET
        url: "http://example.com"
        assertions:
        vars:
          accept:
            from: result.headers.content-type

  - name: UseVar
    steps:
      - type: http
        method: GET
        url: "http://example.com"
        headers:
          test: "{{ .GetVar.accept }}"
        assertions:
          - result.headers.x-cache ShouldEqual "HIT"
      - type: http
        method: GET
        url: "http://example.com"
        headers:
          test: "{{ .GetVar.accept }}"
        assertions:
          - result.headers.accept-ranges ShouldEqual "bytes"
```

This yields the following result:

```
 • Interpolation POC (/tests/interpolation.yml)
 	• GetVar SUCCESS
 	• UseVar FAILURE
Testcase "UseVar", step #1: Assertion "result.headers.acceptµµµranges ShouldEqual \"bytes\"" failed. expected: bytes  got: <nil> (/tests/interpolation.yml:0)
```

A few things to note:
- In the `UseVar` testcase, there are 2 steps defined. They both execute the same HTTP request, but look at different headers from the result.
- The first step succeeds, but the second step fails. Note that the assertion tries to look at the `acceptµµµranges` header, which is incorrect.
- As far as I can tell, this _only_ happens when the test case refers back to a result from a previous test case. If in the example above we replace `{{ .GetVar.accept }}` with `{{ .accept }}` (which is defined in the `vars` key at the top), everything works fine.

In our codebase we use this pattern to execute a login request to generate a JWT token at the beginning of the suite, which we then use as a `Authorization` header in subsequent testcases.

I've tried to debug this a little bit, and was able to trace back the string interpolation to the following function:
https://github.com/ovh/cds/blob/c80557c58fbb2fc28af3da94385172abefddc9fc/sdk/interpolate/interpolate.go#L49

The `Do` function yield different output for both test cases when it's called from: https://github.com/ovh/venom/blob/master/process_testcase.go#L209

I'm not really sure where to go from here, because both test cases have similar raw input, yet a very different `vars` slice. My hunch is that something in the `vars` slice trips up the interpolation function?

fsamin:
closes #475

Signed-off-by: francois  samin <francois.samin@corp.ovh.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ tests.log
 *.jpg
 ./.venomrc
 test_results.xml
+*.prof
--- a/go.mod
+++ b/go.mod
@@ -30,14 +30,14 @@ require (
 	github.com/mitchellh/go-homedir v1.1.0
 	github.com/mitchellh/mapstructure v1.3.3
 	github.com/mndrix/tap-go v0.0.0-20171203230836-629fa407e90b
-	github.com/ovh/cds/sdk/interpolate v0.0.0-20201113080944-f9d9d58c157a
+	github.com/ovh/cds/sdk/interpolate v0.0.0-20211227153055-8d8a5b43e036
 	github.com/ovh/go-ovh v1.1.0
 	github.com/pkg/errors v0.9.1
 	github.com/rubenv/sql-migrate v0.0.0-20200616145509-8d140a17f351
 	github.com/sclevine/agouti v3.0.1-0.20180306165625-6ada53bb069e+incompatible
 	github.com/sijms/go-ora v0.0.0-20201108135513-712ea4f3d160
 	github.com/sirupsen/logrus v1.7.0
-	github.com/spf13/cast v1.3.1
+	github.com/spf13/cast v1.4.1
 	github.com/spf13/cobra v1.1.1
 	github.com/spf13/pflag v1.0.5
 	github.com/streadway/amqp v1.0.0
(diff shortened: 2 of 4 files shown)


ovh-cds:
CDS Report build-venom-a#726.0 ✘  
* Build
  * Build ✔ 
* Tests
  * Acceptance Tests ✔ 
* Analyse
  * Sonar ✘

ovh-cds:
CDS Report build-venom-a#726.1 ✘  
* Build
  * Build ✔ 
* Tests
  * Acceptance Tests ✔ 
* Analyse
  * Sonar ✘

ovh-cds:
CDS Report build-venom-a#726.2 ✘  
* Build
  * Build ✔ 
* Tests
  * Acceptance Tests ✔ 
* Analyse
  * Sonar ✘

richardlt (approved):

Issue #476, opened 2021-12-22
feat: don't log a specific step or test case


nqb:
Hello,

When you use Venom with secrets, you may want to remove secrets from `venom.log`. It could be interesting to have a general setting like `log: False` which can be used to specify that output of a test case or a step should not be wrote into `venom.log`

fsamin:
Hello,

For my understanding are the secret values some variables or are they data coming from testcase ?

Several users would like to hide some data from logs and output. I think about defining a list of variables names to consider as secret. Something like:

```
name: a test suite

secrets:
 - secret-value
 - secret-testcase.myvariable

testcases:
  - name: secret-testcase
    steps:
    - script: echo {{.secret-value}}
      assertions:
      - result.systemout ShouldEquals {{.secret-value}}
     vars:
      myvariable:
        from: result.systemout
```

In this example we assume that secret values will exist in variable `secret-value` (which will be set at runtime) and `secret-testcase.myvariable` (which will be exported during test execution).
It means that venom will hide those values in all logs, output and dump files.

would it be a solution for you needs ?

nqb:
Hello @fsamin,

> For my understanding are the secret values some variables or are they data coming from testcase ?

Secret values are data coming from a test case: we used an `exec` executor to get secrets from an API. Them we assigned values to a variable:

Example: https://github.com/inverse-inc/packetfence/blob/eb3e122db1abef619fabbc041e70d04ad4f6eaee/t/venom/test_suites/configurator/30_run_configurator_step3.yml#L3-L31

> would it be a solution for you needs ?

Looks good to me.

YoannAtOVH:
Hello @fsamin ,

I would prefer something to discard variable loaded from file to appear on log, like secret.
Idea is to replace value output by `<sensitive>`, inspire from [Packer - Sensitive Variables](https://www.packer.io/docs/templates/legacy_json_templates/user-variables#sensitive-variables)

Something like :
`
name: MyTestSuite

vars:
  AK: "wMi3hzKS17lpXB0Jcmg09tdx5SN"

sensitive-variables:
- AK

testcases:
- name: test assertion
  steps:

  - type: http
    method: GET
    url: http://postman-echo.com/get
    headers:
      X-Ovh-Consumer: {{ .AK }}
    timeout: 5

`

Will produce into venom.log like:
`
[INFO] Reading 476.yml
[...]

PR #487, opened 2022-01-12
feat: detailed steps output

### LINKED ISSUE #485: feat request: more detailed output result
It would be nice to have a more detailed output which show all steps, like `range` steps and custom executor `steps`, the current one is pretty barebone and make it hard to know what has been run.

Example:
```diff
$ venom run test/ports.yaml 
• Testsuite
!  • Testcase FAILURE (1/2 tests succeded)
+    • 123 SUCCESS
+      • tcp SUCCESS
+      • udp SUCCESS
!    • 53 FAILURE (1/2 tests succeded)
+      • tcp SUCCESS
-      • udp FAILURE
```

```yaml
# Test suite
testcases:
  - name: Testcase
    steps:
      - type: checkPort
        host: 1.1.1.1
        range: [80, 443]
        port: '{{.value}}'

---
# Custom executor
executor: checkPort
steps:
  - name: {{.value.protocol}}
    type: exec
    script: nc -vz{{.value.arg}} {{.input.host}} {{.input.port}}
    range:
      - protocol: tcp
        arg: ""
      - protocol: udo
        arg: "u"
```

lowlighter:
Closes #485
This is a followup of #453 

This PR introduces a detailed output:
- each step is logged whether it succeeded or not 
  - failures, trace and info are displayed on associated step with correct indentation
  - `range` are expanded and display current key for easier loop tracking
  - `Must` failed assertions display the number of skipped steps 

Also fix some issues:
- `range` step dump were overwritten because they had the same name
- failures in `range` exited the range loop rather than fallthrough (default behaviour of steps)
- `-v` and `` had actually the same verbosity level

Preview (`-v` output):
image

Preview (default output):
image

### DIFF
--- a/process_testcase.go
+++ b/process_testcase.go
@@ -8,6 +8,7 @@ import (
 	"strconv"
 	"strings"
 
+	"github.com/fatih/color"
 	"github.com/ghodss/yaml"
 	"github.com/ovh/cds/sdk/interpolate"
 	"github.com/pkg/errors"
@@ -146,10 +147,11 @@ func (v *Venom) runTestCase(ctx context.Context, ts *TestSuite, tc *TestCase) {
 	}
 
 	defer Info(ctx, "Ending testcase")
-	v.runTestSteps(ctx, tc)
+	v.runTestSteps(ctx, tc, nil)
 }
 
-func (v *Venom) runTestSteps(ctx context.Context, tc *TestCase) {
+func (v *Venom) runTestSteps(ctx context.Context, tc *TestCase, tsIn *TestStepResult) {
+
 	results, err := testConditionalStatement(ctx, tc, tc.Skip, tc.Vars, "skipping testcase %q: %v")
 	if err != nil {
 		Error(ctx, "unable to evaluate \"skip\" assertions: %v", err)
@@ -166,6 +168,7 @@ func (v *Venom) runTestSteps(ctx context.Context, tc *TestCase) {
 
 	var knowExecutors = map[string]struct{}{}
 	var previousStepVars = H{}
+	fromUserExecutor := tsIn != nil
 
 	for stepNumber, rawStep := range tc.RawTestSteps {
 		stepVars := tc.Vars.Clone()
@@ -181,8 +184,10 @@ func (v *Venom) runTestSteps(ctx context.Context, tc *TestCase) {
 		}
 
 		for rangedIndex, rangedData := range ranged.Items {
+			tc.TestStepResults = append(tc.TestStepResults, TestStepResult{})
+			ts := &tc.TestStepResults[len(tc.TestStepResults)-1]
 			if ranged.Enabled {
-				Debug(ctx, "processing step %d", rangedIndex)
+				Debug(ctx, "processing range index: %d", rangedIndex)
 				stepVars.Add("index", rangedIndex)
 				stepVars.Add("key", rangedData.Key)
 				stepVars.Add("value", rangedData.Value)
@@ -225,7 +230,7 @@ func (v *Venom) runTestSteps(ctx context.Context, tc *TestCase) {
 				}
 			}
 
-			Info(ctx, "Step #%d content is: %q", stepNumber, content)
+			Info(ctx, "Step #%d-%d content is: %q", stepNumber, rangedIndex, content)
 			var step TestStep
 			if err := yaml.Unmarshal([]byte(content), &step); err != nil {
 				tc.AppendError(err)
@@ -261,7 +266,10 @@ func (v *Venom) runTestSteps(ctx context.Context, tc *TestCase) {
 				}
 			}
 
-			result := v.RunTestStep(ctx, e, tc, stepNumber, step)
+			printStepName := v.Verbose >= 2 && !fromUserExecutor
+			v.setTestStepName(ts, e, step, &ranged, &rangedData, printStepName)
+
+			result := v.RunTestStep(ctx, e, tc, ts, stepNumber, rangedIndex, step)
 			mapResult := GetExecutorResult(result)
 			previousStepVars.AddAll(H(mapResult))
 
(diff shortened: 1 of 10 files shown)


process_testsuite.go
```diff
@@ -73,6 +73,8 @@ func (v *Venom) runTestCases(ctx context.Context, ts *TestSuite) {
 	var cyan = color.New(color.FgCyan).SprintFunc()
 	var gray = color.New(color.Attribute(90)).SprintFunc()
 
+	// To pass as param
+	detailedOutput := true
```
lowlighter, line 77: - [ ] to check what to do with this

nqb:
Hello @lowlighter,

Thanks for this improvement.

IMHO, we should be careful with display of variables directly in the output because they can contain sensitive informations.
I opened #476 recently and I wonder if it's not better to restrict variables into `venom.log` in place of STDOUT.

cmd/venom/run/cmd.go
```diff
@@ -46,7 +46,7 @@ var (
 	outputDir     string
 	libDir        string
 	stopOnFailure bool
-	verbose       int = 1 // Set the default value for verboseFlag
+	verbose       int
```
lowlighter, line 49: I think this was an error, because running with and without a single `-v` had the same behaviour, so to make it verbose you had to actually put at least `-vv`

In addition, since single `-v` was actually not implemented, imho this is perfect to implement this new report mode

process_testcase.go
```diff
@@ -253,36 +257,43 @@ func (v *Venom) runTestSteps(ctx context.Context, tc *TestCase) {
...
-					failure := newFailure(*tc, stepNumber, "", fmt.Errorf("At least one required assertion failed, skipping remaining steps"))
-					tc.Failures = append(tc.Failures, *failure)
+					failure := newFailure(*tc, stepNumber, rangedIndex, "", fmt.Errorf("At least one required assertion failed, skipping remaining steps"))
+					ts.appendFailure(tc, *failure)
+					v.printTestStepResult(tc, ts, tsIn)
 					return
 				}
-				break
+				v.printTestStepResult(tc, ts, tsIn)
+				continue
```
lowlighter, line 294: `break` replaced by `continue`, else next step weren't executed which contradicts the default behaviour

process_teststep.go
```diff
@@ -62,7 +62,7 @@ func (v *Venom) RunTestStep(ctx context.Context, e ExecutorRunner, tc *TestCase,
 			if oDir == "" {
 				oDir = "."
 			}
-			filename := path.Join(oDir, fmt.Sprintf("%s.%s.step.%d.dump.json", slug.Make(StringVarFromCtx(ctx, "venom.testsuite.shortName")), slug.Make(tc.Name), stepNumber))
+			filename := path.Join(oDir, fmt.Sprintf("%s.%s.step.%d.%d.dump.json", slug.Make(StringVarFromCtx(ctx, "venom.testsuite.shortName")), slug.Make(tc.Name), stepNumber, rangedIndex))
```
lowlighter, line 65: fix steps dump being overwritten on `range`

ovh-cds:
CDS Report build-venom-a#756.0 ✘  
* Build
  * Build ✔ 
* Tests
  * Acceptance Tests ✘

tests/verbose_output.yml
```diff
@@ -0,0 +1,35 @@
...
+  - type: exec
+    script: NO_COLOR=1 {{.venom.executable}} run failing/verbose_output.yml {{.value.opt}}
+    range:
+      verbose:
+        opt: "-v"
+        op: Should
+      default:
+        opt: ""
+        op: ShouldNot
+    assertions:
```
lowlighter, line 16: this test is kind of hacky, but basically it spawns another venom instance (with text colors disable) and run it once in verbose mode and once in non-verbose

then the assertions check that verbose output matches what is intended, wile the non-verbose does not display them

lowlighter:
> IMHO, we should be careful with display of variables directly in the output because they can contain sensitive informations. I opened #476 recently and I wonder if it's not better to restrict variables into `venom.log` in place of STDOUT.

I changed it to only display the only the range key instead. The value output is good in theory, but it can be very long and break the whole output in addition of the issue you mentioned (though I think that looping over secrets is a probably idea in the start)

ovh-cds:
CDS Report build-venom-a#790.0 ✘  
* Build
  * Build ✘

ovh-cds:
CDS Report build-venom-a#798.0 ✘  
* Build
  * Build ✔ 
* Tests
  * Acceptance Tests ✘

cmd/venom/run/cmd.go
yesnault: @lowlighter  with -v, the file venom.log is written.

process_teststep.go
yesnault: ok

tests/verbose_output.yml
yesnault: ok

tests/verbose_output.yml
```diff
@@ -0,0 +1,35 @@
+name: testsuite run in verbose mode
+testcases:
+- name: testsuite run in verbose mode
+  steps:
+  # spawn a venom sub-process and expect it to fail and make assertions on its error messages
+  # ensure no color to avoid annoying checks
+  - type: exec
+    script: NO_COLOR=1 {{.venom.executable}} run failing/verbose_output.yml {{.value.opt}}
```
yesnault, line 8: Can you add something on README about the NO_COLOR please?
In previous release of venom, we removed the color as it was not working on some environment (NO_COLOR did not exist at this time). We can re-add them, but need to provide a way to disable them, NO_COLOR is perfect for that.

yesnault (changes requested):

cmd/venom/run/cmd.go
fsamin: i remember this bug fix, the default value is 1 so we should let the initialization to 1

fsamin (changes requested):

tests/verbose_output.yml
lowlighter: ok, I added it
Found it in https://github.com/fatih/color#disableenable-color for reference (the underlying lib used for logs)

yesnault (approved):

ovh-cds:
CDS Report build-venom-a#892.0 ✘  
* Build
  * Build ✔ 
* Tests
  * Acceptance Tests ✘

fsamin (approved):

yesnault (approved):

CozThatsWhy:
I noticed a formatting issue with the output when verbose is set to "1"
```
• Test Verbose (endpoints/verbose.yml)
 	• 200-Response-Valid-Content-Type 		• http PASS
 PASS
 	• 404-Response-Invalid-Content-Type 		• http PASS
 PASS 
 ```
Verbose 2 looks a bit better:
```
• Test Verbose (endpoints/verbose.yml)
 	• 200-Response-Valid-Content-Type
 		• http PASS
	  	  [trac] writing ***
 	• 404-Response-Invalid-Content-Type
 		• http PASS
	  	  [trac] writing ***
```
Also, maybe a nice QOL feature is to print out the step name if the variable 'name' has a value to it, regardless of verbosity.  This would especially be useful for a range of endpoints in a single executor (without the range=X) value.  If the verbosity is set, then possibly a format of `{{name}} [{{executor}} (range=X)]`

Thanks!

PR #500, opened 2022-01-28
Gitpod integration to Venom

scraly:
In order to facilitate the possibility of contributing to Venom, without having to configure the setup on your computer, with everything ready to use, this PR adds integration to GitPod.

GitPod is an Open Source IDEaaS, surprisingly simple and effective solution The principle? A cloud-based development environment, based on Git, Eclipse Theia and Docker, allowing you to develop, test and deploy your code right from a browser window.

If you want to know more about Gitpod:
https://www.youtube.com/watch?v=pDu_rTh_s-0

With this feature, contributors will no longer need to:
- install Go in their computer
- use a Go correct version or update to a good one
- know the command to start the project

To contribute, they will only need to click on "Open to GitPod"... and start to contribute :).

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,6 @@
+tasks:
+  - init: go get -v -t -d ./...
+    command: go run cmd/venom/main.go -h
+vscode:
+  extensions:
+    - golang.go
--- a/README.md
+++ b/README.md
@@ -7,8 +7,6 @@ Venom is a CLI (Command Line Interface) that aim to create, manage and run your 
 [![Go Report Card](https://goreportcard.com/badge/github.com/ovh/venom)](https://goreportcard.com/report/github.com/ovh/venom)
 [![Dicussions](https://img.shields.io/badge/Discussions-OVHcloud-brightgreen)](https://github.com/ovh/venom/discussions)
 
-[![Open in Gitpod](https://gitpod.io/button/open-in-gitpod.svg)](https://gitpod.io/#https://github.com/ovh/venom.git)
-
 # Table of content
 
 * [Overview](#overview)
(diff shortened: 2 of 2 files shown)


ovh-cds:
CDS Report build-venom-a#811.0 ✘  
* Build
  * Build ✔ 
* Tests
  * Acceptance Tests ✘

.gitpod.yml
```diff
@@ -0,0 +1,6 @@
+tasks:
+  - init: go get -v -t -d ./...
```
yesnault, line 2: Is it possible to get the latest release ?
I'm not a fan of launching venom by bypassing the entire CI. Is it possible to run something like that:

```
LAST_RELEASE=$(curl -s https://api.github.com/repos/ovh/venom/releases | grep tag_name | head -n 1 | cut -d '"' -f 4) && \
DL=$(curl -s https://api.github.com/repos/ovh/venom/releases | grep ${LAST_RELEASE} | grep browser_download_url | grep -E 'linux-amd64' | cut -d '"' -f 4) &&
curl -L $DL -o venom
```

yesnault (changes requested):

.gitpod.yml
```diff
@@ -0,0 +1,6 @@
...
+    command: go run cmd/venom/main.go -h
```
fsamin, line 3: If the goal of this integration is to have dev environment for dev, i suggest to try to use the Makefile to build and maybe run the tests ?

scraly: The goal is to have an IDE in the cloud/in the browser. But not only an IDE but a complete dev environment.
No need to install, in your local machine(s), the correct version of Go, all the tools, the extensions in VSCode... it's one of the power of GitPod.

fsamin: Great ! 
So I guess we should try to only relies on the Makefile
We don't need the `go get` and `make build` should be fine

scraly: When I am developing an application in Go, I don't want to build my binary but to make a go run in order to test in live my changes, for example.
I just want to have my go dependencies in order to then execute `go run` command.

fsamin:
I am closing this because it has been merged in #559

PR #504, opened 2022-02-03
Fix line numbers being duplicated

seanlane:
Noticed that line numbers appear to be duplicated. I believe it stems from these lines. If `linenumber > 0` then the line number statement will be added twice, at lines 85 and 92 of the previous commit. This moves the second conditional statement into the else-clause of the first conditional statement, so it's only added once.

Signed-off-by: Sean Lane <sean.lane@neighbor.com>


yesnault (approved):

fsamin (approved):

fsamin:
Thank you for your contribution 👍

PR #517, opened 2022-03-10
Add feature to produce messages using latest schema available version

sixstone-qq:
This way there is no need to provide the AVSC file upon every message by retrieving the latest version from the schema registry, if available.

Code is using a [no longer maintained library](https://github.com/lensesio/schema-registry), I'd suggest to move to https://github.com/riferrei/srclient which is actively maintained.

If you agree, I can open a new PR with that change :).

Thanks for this great project!


ovh-cds:
CDS Report build-venom-a#835.0 ✘  
* Build
  * Build ✔ 
* Tests
  * Acceptance Tests ✘

fsamin (approved):

PR #550, opened 2022-05-10
handle param like charset

guoweis-outreach:
this is related to https://github.com/ovh/venom/pull/528

sometimes there is extra param after the content type, like below

```
Content-Type: application/vnd.api+json; charset=utf-8
```


yesnault (approved):

fsamin:
Thanks for your contribution
LGTM

PR #553, opened 2022-06-05
Cache executor files

philippgille:
Hello 👋 ,

We're running Venom inside of a Docker container in some cases, while mounting the scenario and custom executor files from the host into the container. Depending on the OS and the Docker config, for example when on macOS and having the "new Virtualization framework" enabled, but "VirtioFS" disabled, file operations can be very slow in Docker. Now with Venom we noticed that custom executors are read dozens of times during scenario execution, adding up to a lot of time.

Here's a screenshot of an example problematic Docker configuration on macOS:
image

But even with both of those options disabled, or both enabled, the file I/O is not great. For example we have a scenario that takes 1m, and 10s of that is spent purely on reading custom executor files!

This PR fixes that.

It introduces an executor file cache, and reads each file only _once_.

Cc @christophe-dufour 

PS: Maybe some other operations in the loop don't need to be repeated as well, but of those, they either don't take long (based on my measurements) and an optimization would have little impact, or the interpolation step (where an optimization would be nice) is a bit more complex to optimize, because for each `registerUserExecutors` call there can be new variable values so the interpolated result can't be cached, if my understanding is correct. And I wanted to keep this PR simple.

### DIFF
--- a/venom.go
+++ b/venom.go
@@ -32,13 +32,14 @@ type ContextKey string
 // New instanciates a new venom on venom run cmd
 func New() *Venom {
 	v := &Venom{
-		LogOutput:        os.Stdout,
-		PrintFunc:        fmt.Printf,
-		executorsBuiltin: map[string]Executor{},
-		executorsPlugin:  map[string]Executor{},
-		executorsUser:    map[string]Executor{},
-		variables:        map[string]interface{}{},
-		OutputFormat:     "xml",
+		LogOutput:         os.Stdout,
+		PrintFunc:         fmt.Printf,
+		executorsBuiltin:  map[string]Executor{},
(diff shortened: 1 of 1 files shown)


yesnault (approved):
LGTM, thank you!

christophe-dufour:
Hey @yesnault 

Thanks for approving the pull request :) 

I've 2 questions:
- Do you know when it will be merged?
- Do you know when the next venom release will be?

We are using Venom a lot and that PR is bringing huge improvement

ovh-cds:
CDS Report build-venom-a#883.0 ✘  
* Build
  * Build ✔ 
* Tests
  * Acceptance Tests ✘

christophe-dufour:
Awesome news. Thx for the update @yesnault

PR #570, opened 2022-08-12
feat: VENOM_PRESERVE_CASE default ON

yesnault:
It's now possible to read something like that :

```json
{
  "foo": "bar",
  "FOO": "truc",
}
```
and use these variables:

```yml
- type: readfile
    path: the_file.json
    info: "{{.Result.ContentJSON.foo}}-{{.Result.ContentJSON.FOO}}"
```

Environment variable: VENOM_PRESERVE_CASE

venom 1.1.x
  VENOM_PRESERVE_CASE="AUTO" is equals to VENOM_PRESERVE_CASE="OFF"

venom 1.2.x
  VENOM_PRESERVE_CASE="AUTO" is equals to VENOM_PRESERVE_CASE="ON"

venom 1.1 users, you can try:

```sh
> export VENOM_PRESERVE_CASE="ON";
> venom run ...
```

### DIFF
--- a/dump.go
+++ b/dump.go
@@ -12,7 +12,7 @@ var preserveCase string
 func init() {
 	preserveCase = os.Getenv("VENOM_PRESERVE_CASE")
 	if preserveCase == "" || preserveCase == "AUTO" {
-		preserveCase = "OFF"
+		preserveCase = "ON"
 	}
 }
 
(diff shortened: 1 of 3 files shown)


kamil-muszynski-90poe:
Hello, I came across this PR after trying to figure out why I can't access my (camel-cased) json keys when asserting on result.systemoutjson.key... Maybe - as a part of this PR - it would be a good idea to mention in the documentation that keys are converted to lowercase by default?

fsamin (approved):

PR #573, opened 2022-08-21
Change the default verbosity to 0

samj1912:
venom's default verbosity level should be 0 unless set by the user. This follows the general unix philosophy of tools being quiet unless there is an error


yesnault:
already in https://github.com/ovh/venom/pull/572

PR #584, opened 2022-09-02
Fix waitFor and messagesLimit behaviour in Kafka consumer

sixstone-qq:
This PR includes two fixes over Kafka consumer executor:

1. waitFor didn't work in some cases mixed with timeout property
   as it is reused by test executor. Using a goroutine over
   consume claim we can work on timeout and wait for without issues.

2. messageLimit after #448 may only work if there are more messages to
   consume in the topic, if the message limit is hit after consuming the
   last message from the topic, the condition is never met and it waits
   until timeout is reached leading to longer Kafka consumer step
   executions.

### DIFF
--- a/executors/kafka/kafka.go
+++ b/executors/kafka/kafka.go
@@ -229,11 +229,11 @@ func (e Executor) getMessageValue(m *Message, workdir string) ([]byte, error) {
 	// This is test with Avro
 	var (
 		schemaID int
-		schema string
+		schema   string
 	)
 	// 1. Get schema with its ID
 	// 1.1 Try with the file, if provided
-	subject := fmt.Sprintf("%s-value", m.Topic)  // Using topic name strategy
+	subject := fmt.Sprintf("%s-value", m.Topic) // Using topic name strategy
 	schemaFile := strings.Trim(m.AvroSchemaFile, " ")
 	if len(schemaFile) != 0 {
 		schemaPath := path.Join(workdir, schemaFile)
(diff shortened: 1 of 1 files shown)


philippgille (approved):
I forgot the second msg limit check in the other PR. Thanks for fixing it! LGTM

yesnault (approved):
LGTM, thank you.

fabriziomoscon:
I am using: `v1.1.0-rc.1`

@sixstone-qq @philippgille I am experiencing that `waitFor` is now mandatory for kafka consumer executor. (note that the readme says optional).
If I don't specify it I receive a `context deadline exceeded` error:
```
step #2-0: Assertion "result.err ShouldBeEmpty" failed. expected 'kafka consumed failed: context deadline exceeded'
```

my full consumer looks as follow:

```yaml
- type: kafka
      clientType: consumer
      withTLS: false
      withSASL: false
      withAVRO: true
      kafkaVersion: "2.6.0"
      markOffset: true
      initialOffset: oldest
      waitFor: 1
      groupID: venom
      addrs:
        - "{{.kafkaHost}}:{{.kafkaPort}}"
      schemaRegistryAddr: "{{.kafkaSchemaRegistryHost}}"
      topics:
        - "{{.ticketUpsertCommand}}"
      assertions:
        - result.err ShouldBeEmpty
        - result.messages.__len__ ShouldEqual 1
        - result.messagesjson.messagesjson0.value ShouldNotBeEmpty
        - result.messagesjson.messagesjson0.value.description ShouldContainSubstring "Strong Winds in 2 hours"
```

PR #588, opened 2022-09-12
feat: improve output with range

### LINKED ISSUE #583: Dynamic Test Case Generation
We are working on transitioning from our current Mocha implementation of tests to venom, and one of the things we perform is a `for each` loop with our endpoints that then performs a test case.  This way when we add more endpoints, we can add to the array of endpoints to test.  

We tried some functionality with Venom, but saw that if we are using a ranged value in a single step will only print the name of the test case and it's status (with a ranged=X) if the verbosity is set to > 1.  Setting it to 0 will not print the step at all, and only print the testcase status.

Is there a way that we can generate test cases off a range so we can write one test case but it is created via a range?  Or maybe be able to print out the steps of a test case in 0 Verbosity if it is a range (and without the range=X).

Thank you!

yesnault:
close #583

### DIFF
--- a/process_testcase.go
+++ b/process_testcase.go
@@ -268,6 +268,7 @@ func (v *Venom) runTestSteps(ctx context.Context, tc *TestCase, tsIn *TestStepRe
 
 			tsResult.Number = stepNumber
 			tsResult.RangedIndex = rangedIndex
+			tsResult.RangedEnable = ranged.Enabled
 			tsResult.InputVars = vars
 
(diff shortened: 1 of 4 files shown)


ovh-cds:
CDS Report build-venom-a#1026.0 ✘  
* Build
  * Build ✔ 
  * Unit Tests ✔ 
* Tests
  * Acceptance Tests ✘

fsamin (approved):

PR #592, opened 2022-10-17
feat: `skip` support at step level

### LINKED ISSUE #486: feat request: `skip` at step level
If I'm not mistaken it's not possible to use `skip` at step level, it would be nice to be able to at least skip some tests in custom executors.

The main objective of custom executors (I think) is to reduce and factorize tests description, but sometimes you just need to skip out a step in a few case but it seems overkill to duplicate all steps and create another executor just to achieve the same 

Example:
```yaml
executor: test
input: 
  script1: true
  script2: true
steps:
  - type: exec
    script: script1
    skip:
      - {{.input.script1}} SouldEqual true
  - type: exec
    script: script2
    skip:
      - {{.input.script2}} SouldEqual true
```

lowlighter:
Closes #486 

Make it possible to `skip` at step level (currently it's only possible to do so at case level)
If all steps from a case are skipped, then the case will also change its state to "SKIP" rather than "PASS"

*Default output:*
image

*Verbose output:*
image

Additionally this slightly changes the output of skipped test case when verbose mode is enabled to display that "all steps were skipped" rather than nothing, which could be misleading as a user may think that something is wrong with their test case definition

<details>
<summary>Current output</summary>

image

</details>

### DIFF
--- a/process_testcase.go
+++ b/process_testcase.go
@@ -302,24 +302,31 @@ func (v *Venom) runTestSteps(ctx context.Context, tc *TestCase, tsIn *TestStepRe
 			printStepName := v.Verbose >= 1 && !fromUserExecutor
 			v.setTestStepName(tsResult, e, step, &ranged, &rangedData, rangedIndex, printStepName)
 
-			tsResult.Start = time.Now()
-
 			// ##### RUN Test Step Here
-			tsResult.Status = StatusRun
-			result := v.RunTestStep(ctx, e, tc, tsResult, stepNumber, rangedIndex, step)
-			if len(tsResult.Errors) > 0 || !tsResult.AssertionsApplied.OK {
+			skip, err := parseSkip(ctx, tc, tsResult, rawStep, stepNumber)
+			if err != nil {
+				tsResult.appendError(err)
 				tsResult.Status = StatusFail
+			} else if skip {
+				tsResult.Status = StatusSkip
 			} else {
-				tsResult.Status = StatusPass
+				tsResult.Start = time.Now()
+				tsResult.Status = StatusRun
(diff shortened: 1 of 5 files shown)


process_testcase.go
```diff
@@ -302,24 +302,31 @@ func (v *Venom) runTestSteps(ctx context.Context, tc *TestCase, tsIn *TestStepRe
 			printStepName := v.Verbose >= 1 && !fromUserExecutor
 			v.setTestStepName(tsResult, e, step, &ranged, &rangedData, rangedIndex, printStepName)
```
lowlighter, line 304: Diff below looks complicated but it actually just move all the code related to `v.RunTestStep` to the `else` block of the `if skip` conditional

process_testsuite.go
```diff
@@ -117,13 +117,17 @@ func (v *Venom) runTestCases(ctx context.Context, ts *TestSuite) {
...
 			}
 			if testStepResult.Status == StatusFail {
 				hasFailure = true
 			}
+			if testStepResult.Status == StatusSkip {
+				skippedSteps++
+			}
 		}
 
 		if verboseReport || hasRanged {
```
lowlighter, line 133: Not sure if it's because of my environment, but the newline seems to be printed too late when I run venom in verbose mode:
image

In my tests I moved this block to line 114 (just before `v.runTestCase(ctx, ts, tc)`) and it seems to yield the correct output.

Since the output from https://github.com/ovh/venom/issues/583#issuecomment-1243679828 looked correct, I didn't touched it, but just wanted to make a note about it

fsamin (approved):

process_testsuite.go
yesnault: you're right, fixed.

yesnault (approved):

PR #615, opened 2023-01-14
fix: avoid `nil` pointer dereference in `ShouldContain` assertions

iamleot:
It is possible to trigger `nil` pointer dereference by calling `ShouldContain` assertions.

Possible minimal reproducer in `venom`:

```yaml
testcases:
  - steps:
      - assertions:
          - result.systemoutjson ShouldContain value
```

This PR fixes that by avoiding unconditional calls to `Kind()` when `ShouldContain()` contains a `nil` pointer like in previous example and adds a test to double-check that.

### DIFF
--- a/assertions/assertions.go
+++ b/assertions/assertions.go
@@ -549,7 +549,7 @@ func ShouldContain(actual interface{}, expected ...interface{}) error {
 	if err := need(1, expected); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


fsamin (approved):
thanks for your PR
LGTM

yesnault (approved):

iamleot:
Thank you!

PR #617, opened 2023-01-27
feat: removed result from testcase variables and removed errors warning and info from non verbose runs

### LINKED ISSUE #595: Improving error debugging (especially in CI)
Hello folks! :)

The `info` keyword is great, but it always print (even if the test is successful).

It would be very useful to have something that only prints on failure. So that when everything is fine, it does not pollute the output and in case of an error, you have enough info to debug (it's especially true when a test fails in a CI pipeline and only in CI).

Currently the workaround is to perform an "dummy" assertion on systemout/systemerr (result.systemout ShouldEqual "") so that when it fails, you get the output.

At first, I was thinking of the new keyword. But maybe a new verbosity level that prints the all `result` in case of a failure would be simpler and a lot more convenient. 

What do you think? I can try to work on a PR :)

On another note: I've been using venom since I joined OVHcloud a few months ago, and I like it very much. It's a great little piece of software :)

### LINKED ISSUE #597: Print info in current "step" before assertions
In this case, `result 2` will still print the bodyjson of `result 1`. 
Is there a way to make `result 2` print the bodyjson of `test2`? 

Hope the `result 2` will be printed even if the test of `test2` failed.

```yaml
testcases:
- name: test A
  steps:
  - type: http
    method: POST
    url: {{.url}}/test1
    timeout: 5
    info: result 1 {{.result.bodyjson}}
    assertions:
  - type: http
    method: GET
    url: {{.url}}/test2
    timeout: 5
    retry: 3
    delay: 2
    info: result 2 {{.result.bodyjson}}
    assertions:
```

kilianpaquier:
- removed ranged prints on non-verbose runs
- fixed info not interpolating right variable in step X with .result variable (was showing variable from X-1)
- print info before errors: closes #597 
- on non-verbose runs, print info only when there're failures: closes #595

### DIFF
--- a/process_testcase.go
+++ b/process_testcase.go
@@ -312,7 +312,7 @@ func (v *Venom) runTestSteps(ctx context.Context, tc *TestCase, tsIn *TestStepRe
 			} else {
 				tsResult.Start = time.Now()
 				tsResult.Status = StatusRun
-				result := v.RunTestStep(ctx, e, tc, tsResult, stepNumber, rangedIndex, step)
+				v.RunTestStep(ctx, e, tc, tsResult, stepNumber, rangedIndex, step)
 				if len(tsResult.Errors) > 0 || !tsResult.AssertionsApplied.OK {
 					tsResult.Status = StatusFail
 				} else {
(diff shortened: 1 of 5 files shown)


yesnault:
revert. Intergration Tests are failing, example: `venom run  interpolate_once.yml`

kilianpaquier:
> revert. Intergration Tests are failing, example: `venom run interpolate_once.yml`

Yes sorry, didn't see that one. It doesn't work anymore because I removed all .result from testcase variables.

Only the assigned variables (coming from the key 'vars') are present in testcase variables, that's why interpolation.yaml doesn't fail.

In my opinion, .result should be teststep scoped. If .result must be accessed later, then it should be assigned to a variable during the teststep. It is because every step computes a .result and if multiple steps are to compute .result.systemout, all of them would be stored in the testcase variables but only the last one would be used.

In this case, result.systemout should be assign to a variable, and the comparison should be done with it. Just like it's done in interpolation.yaml.

Another possibility is to prefix .result with the teststep name, like it is done on variable in different testcases.

Issue #604, opened 2022-11-22
starting process as daemon, leaving running through all steps of test


amlwwalker:
I have a series of steps. In the first step I start a daemon process - I'm using Go, so I'm doing something along the lines of

```go
	exec.Command("some-command", "&")
	cmd.Start()
```

The first step starts the daemon and runs some tests against it (its a web service). Then in each `step` of the test I am running different tests against the same daemonised instance. Its important that I am using the same instance.

The issue is, that venom won't move to the next step in the test because the process is still running even though I daemonised it. Is there a way to leave it running throughout the series of steps when its started in the first step?

I am aware of being able to do

```
	cmd := exec.Command(command, arguments...)
	fmt.Println(cmd.String())
	if err := cmd.Start(); err != nil {
		return err
	}
	err := cmd.Process.Release()
	if err != nil {
		fmt.Println("cmd.Process.Release failed: ", err)
		return err
	}
```

however for some reason, even detaching from the process doesn't allow venom to move on and it continues to hang before moving to the next step.

I would start the process totally seperately, although part of the starting of the process relies on me generating some files, which is part of the test. So in essence I am using venom to do:

- generate required startup files
- start process with previously generated files and detach from it
- fire requests against the detached process
- close the process down

I can detach from within my Go application, but for whatever reason venom won't detach from it?

fsamin:
Hello,

Can you elaborate a bit more ? you want venom to start your background application ?
According to https://github.com/ovh/venom/blob/master/executors/exec/exec.go#L191, it waits for the end of the process.

If you need to run your tests against a background application, you can start it before running venom.

fehrnah:
I am in the same situation.

I need to start a long running process with parameters depending on previous steps of a test.
Ideally, this would be a different executor than exec, something like run.

I would expect the executor to have these two actions:
- start: take a command line, use it to start a background process and return a "handle" to it.
- stop: take a handle,  use it to stop the process and return stdout, stderr and return code.

fsamin:
If you need to start a long running process to run your tests, venom should not be responsible to orchestrate this. It only run tests.
The most recurrent usage i know is for instance to start an application and its dependencies with docker-compose and then to run venom tests against it.

BTW if you needs something specific it's possible to build an executor as a plugin.
Here is an example of a plugin https://github.com/ovh/venom/tree/master/executors/plugins/odbc

fehrnah:
In case this helps someone else, here is how I managed to do it : 

```yaml
[...]

PR #623, opened 2023-02-08
fix(exec): panic on cmd.Start error

fehrnah:
When trying to do **crimes** to get around  #604 I found a way to crash the exec executor.

This fixes said crash. 

Sadly I don't have the code that caused the error anymore so I can't provide it here.


fsamin (approved):
thanks for your contribution
LGTM

PR #627, opened 2023-02-10
fix: TestSuiteXML is missing the Time property

ivan-velasco:
Value is converted from float64 to string. This will resolve gh actions test reporters for junit xml, showing missing test suite time execution values

Screenshot 2023-02-09 at 7 54 22 PM


yesnault:
LGTM - Notice that ts.Duration contains value in Seconds, not ms.

yesnault (approved):

PR #632, opened 2023-02-10
feat: allow overriding default `workdir`

ivan-velasco:
when working with custom http executors, I have a need to use a directory of common reusable json payloads defined in `{{.venom.libdir}}`

If this looks ok, I will provide test and documentation.

```yaml
  workdir: "{{.venom.libdir}}"
  bodyfile: "/testdata/test.json"
```

Signed-off-by: Ivan Velasco <ivan.velasco@socotra.com>


fsamin:
I don't like this "workdir" option on the http executor.
I just push this PR: https://github.com/ovh/venom/pull/635
In order to solve your issue i think you should be able to do :

```
  bodyfile: "{{.venom.libdir}}/testdata/test.json"
```

with `{{.venom.libdir}}` set as an absolute file

ivan-velasco:
This works much better thanks @fsamin

PR #653, opened 2023-04-13
feat(mongo): add "loadFixtures" action

Thiht:
It works roughly the same as the dbfixtures executor. I wanted to implement it directly in dbfixtures but it's too tied to database/sql so it doesn't work well with Mongo

Note that I used yaml.v3 instead of v2, because v2 unmarshals yaml objects as `map[any]any` instead of `map[string]any`, it's too hard to deal with when inserting in Mongo

### DIFF
--- a/executors/mongo/README.md
+++ b/executors/mongo/README.md
@@ -19,6 +19,33 @@ Example:
 }
 ```
 
+### Load fixtures
+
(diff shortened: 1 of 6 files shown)


yesnault (approved):
Thank you!

PR #667, opened 2023-05-12
Adding secrets field and refactoring the logging process. 

fokion:
Adding secrets in the context and creating a function that we can call to hide sensitive data

Add the ability to mark variables as secrets in the testsuite so that the venom.log does not show them

#476


fokion:
Is this approach on the right track @yesnault  , @fsamin  ?

fokion:
Hey @yesnault  , @fsamin  is there an eta for merging this PR?

PR #668, opened 2023-05-17
feat: Logging level flag added

ivan-velasco:
- feature: to partially solve this issue, introduced `loggingLevel` 
-  supported values are based on logrus package
```
"panic"
"fatal"
"error"
"warn"
"info"
"debug"
"trace"
```
https://github.com/ovh/venom/issues/666

Additional discussion should occur as this logging package is in maintenance mode and performance is not the best. Migrate to a different library in the future. For instance https://github.com/rs/zerolog


fokion:
👍

Issue #633, opened 2023-02-13
Add new JSON equality operators


Thiht:
Hi!

I propose to add 4 new assertion operators: `ShouldEqualJSON`, `ShouldEqualJSONFile`, `ShouldContainJSON`, `ShouldContainJSONFile`.

These operators work on JSON, so they ignore whitespace, key ordering, [comments](https://json5.org/#comments) and other no-ops.

- `ShouldEqualJSON` asserts that 2 JSON documents are equal, according to the above rule.

```
// TRUE
{
  "bar": 1,
  "foo": true
} ShouldEqualJSON { "foo": true, "bar": 1 }
```

- `ShouldEqualJSONFile`: same as `ShouldEqualJSON`, but reads `expected` from a file.

```
result.body ShouldEqualJSONFile ./expected.json
```

I mentioned allowing JSON comments above (supported in [JSON5](https://json5.org/) because it would be extremely useful with files, to be allowed to add details on some expected values.

Now in my opinion, 2 more interesting operators are `ShouldContainJSON` and `ShouldContainJSONFile`. They would work the same as `ShouldEqualJSON`, `ShouldEqualJSONFile`, except that they could be used in all cases where some keys are unpredictable: random values, generated IDs (UUIDs, etc.), timestamps... Example:

```
// TRUE
{ "id": "63e60deaf57eda4b4e4c511e", "timestamp": 1676311874, "status": "ok" } ShouldContainJSON { "status": "ok" }

// We can still add more assertions below if we want
result.bodyjson.id ShouldNotBeEmpty
result.bodyjson.timestamp ShouldBeGreaterThan 0
```

What do you think? I think `ShouldContainJSON`/`ShouldContainJSONFile` are a bit more challenging to implement, but would be extremely useful. They would help get rid of this situation when handling big JSONs:

```
result.bodyjson.a ShouldEqual xxx
result.bodyjson.b ShouldEqual xxx
result.bodyjson.c ShouldEqual xxx
// ...
result.bodyjson.x ShouldEqual xxx
result.bodyjson.y ShouldEqual xxx
result.bodyjson.z ShouldEqual xxx
```

Edit: I just found this lib which does exactly that, so no need to implement diffing algorithms: https://pkg.go.dev/github.com/nsf/jsondiff#section-readme

> At the moment it can detect a couple of types of differences:
> - FullMatch - means items are identical.
> - SupersetMatch - means first item is a superset of a second item.
> - NoMatch - means objects are different.

[...]

PR #676, opened 2023-06-01
Add ShouldJSONEqual assertion

philippgille:
This implements `ShouldJSONEqual`, which is one of the requested features in https://github.com/ovh/venom/issues/633.

The naming can also be `ShouldEqualJSON`, but with `ShouldTimeEqual` already existing, the other one seems to be more consistent with that pattern.

### DIFF
--- a/README.md
+++ b/README.md
@@ -695,7 +695,7 @@ Open it in your favorite editor or IDE and fill it with this content:
 name: APIIntegrationTest
 
 vars:
-  url: https://eu.api.ovh.com/
+  url: https://eu.api.ovh.com
 
 testcases:
 - name: GET http testcase, with 5 seconds timeout


philippgille:
Sorry for the forced push, I forgot the commit sign-off again for the last two commits

philippgille:
@yesnault is there room in v1.2.0 for this, or too late?

yesnault:
> @yesnault is there room in v1.2.0 for this, or too late?

Hi @philippgille , thank you for this PR, it's a great new feature.

No problem, it could be integrated into the 1.2.0 version.

assertions/assertions.go
```diff
@@ -1174,6 +1174,120 @@ func ShouldTimeEqual(actual interface{}, expected ...interface{}) error {
...
+	if err != nil {
+		return err
+	}
+
+	// `actual` can be a map, slice or the string representation of a JSON object or array.
+	var actualMap map[string]interface{}
+	var actualSlice []interface{}
+	var obj bool
+	switch actual.(type) {
+	case map[string]interface{}:
```
yesnault, line 1208: I think that the parser to detect the json could be more json compliant. There is a check for `{` or `[`, but a simple text or number is also a valid number, see all values on  https://www.json.org/json-en.html

yesnault (changes requested):

assertions/assertions.go
yesnault: I prefer the first proposal: let the choice to the user and having the ShouldJSONEqual supporting the different types. Maybe not the easiest to implement, but definitely the easiest for users to understand if we say that ShouldJSONEqual supports the full JSON spec.

philippgille: Makes sense 👍 

Done in 5c05103a6d3b69b66ae3cde5e374f4f7a5e57a05

⚠️ JSON `null` are tricky, because Venom passes it as empty string and not Go's `nil`. Haven't checked if this is a fundamental limitation in the upstream callers in Venom, or if it can be changed. Do you know maybe?

And added to README in ba3769c72a0ba06a0a0daaaaf8fd663c27fe473f

Looking at the assertion list in the README it's a bit messy, I think it would make sense to sort the assertions alphabetically. But then the `ShouldJSONEqual` feels misplaced. => WDYT about naming it `ShouldEqualJSON` like originally proposed in #633, and then for consistency create `ShouldEqualTime` as an alias for `ShouldTimeEqual`, which is required for backward compatibility. The latter could be marked as deprecated in the README.

yesnault: About naming, `ShouldEqualJSON ` is ok too, perhaps we can add some alias in another pr than here if needed. Thank you for adding diffrents json values.
LGTM

yesnault (approved):

PR #704, opened 2023-07-18
reporting optimisation, generate the reports as we execute the tests

fokion:
#701 

Creating the reports as we execute the tests and clear them from memory and generate the html report from those in the end. Let me know your thoughts  @yesnault @fsamin


fokion:
closing this as it is part of https://github.com/ovh/venom/pull/708

PR #713, opened 2023-08-25
Feature/default http headers

ivan-velasco:
Enhance header handling efficiency

In situations where a significant number of test cases require the same headers to be consistently set, the existing approach led to redundancy. Prior to this enhancement, header definitions were repeated for each request, as exemplified:

```yaml
- type: http
  method: PUT
  headers:
    Authorization: ***
    Accept: application/json
    Content-Type: application/json
```
The objective of this feature is to streamline the process by reducing repetitive lines for the commonly used content type. Subsequently, the optimized configuration appears as follows:
after
```yml
    - type: http
      method: PUT
      headers:
        Authorization: ***
```

Furthermore, it's important to note that if the context involves `multipart_form` requests, no default headers will be automatically set.

### DIFF
--- a/cmd/venom/run/cmd.go
+++ b/cmd/venom/run/cmd.go
@@ -180,6 +180,11 @@ func initFromReaderConfigFile(reader io.Reader) error {
 	}
 	if configFileData.LibDir != nil {
 		libDir = *configFileData.LibDir
+		if absPath, err := filepath.Abs(libDir); err == nil {
+			libDir = absPath
+		} else {
+			return err
+		}
 	}
 	if configFileData.OutputDir != nil {
 		outputDir = *configFileData.OutputDir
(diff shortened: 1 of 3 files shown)


yesnault:
It really depends on the purpose of the test. Perhaps user want to check the behavior of the APi without any header. I prefer to not add default value for headers and let user to use an UserExecutor for that.

ivan-velasco:
@yesnault Thank you for sharing your thoughts! I appreciate your perspective on the default headers and understand that they might not be suitable for every use case. 

Adding default header into the http executor significantly reduced the codebase in my case. If we need to bypass the default headers we can add a flag in configuration or set `Content-Type` to `nil`
Incorporating a UserExecutor for every HTTP executor may also lead to excessive complexity.

If you have any more suggestions or thoughts on how we can further enhance http executor, including using a common auth token without having to specific in each request that would be great.

PR #721, opened 2023-09-23
fix(venom): env variables with equal sign

bramca:
This PR fixes the following issue:
VENOM_VAR_a='host=localhost'
results in a='host' because of the split on '=' and only taking the first value back

With this the var would end up like this:
a = 'host=localhost'

Making it possible to add equal signs in the variable

### DIFF
--- a/cmd/venom/run/cmd.go
+++ b/cmd/venom/run/cmd.go
@@ -289,7 +289,7 @@ func initFromEnv(environ []string) ([]string, error) {
 		if strings.HasPrefix(env, "VENOM_VAR_") {
 			tuple := strings.Split(env, "=")
 			k := strings.TrimPrefix(tuple[0], "VENOM_VAR_")
-			variables = append(variables, fmt.Sprintf("%v=%v", k, cast(tuple[1])))
+			variables = append(variables, fmt.Sprintf("%v=%v", k, cast(strings.Join(tuple[1:], "="))))
 		}
 	}
 
--- a/cmd/venom/run/cmd.go
+++ b/cmd/venom/run/cmd.go
@@ -287,9 +287,9 @@ func initFromEnv(environ []string) ([]string, error) {
 
 	for _, env := range environ {
 		if strings.HasPrefix(env, "VENOM_VAR_") {
-			tuple := strings.Split(env, "=")
+			tuple := strings.SplitN(env, "=", 2)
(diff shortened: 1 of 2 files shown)


yesnault (changes requested):
Hi @bramca , thank you for your proposal.

Can you add a unit test about that into https://github.com/ovh/venom/blob/master/cmd/venom/run/cmd_test.go#L92 please?

bramca:
Added a unit test for it

cmd/venom/run/cmd.go
```diff
@@ -289,7 +289,7 @@ func initFromEnv(environ []string) ([]string, error) {
 		if strings.HasPrefix(env, "VENOM_VAR_") {
 			tuple := strings.Split(env, "=")
```
yesnault, line 290:
```suggestion
			tuple := strings.SplitN(a, "=", 2)
```

cmd/venom/run/cmd.go
```diff
@@ -289,7 +289,7 @@ func initFromEnv(environ []string) ([]string, error) {
...
 			k := strings.TrimPrefix(tuple[0], "VENOM_VAR_")
-			variables = append(variables, fmt.Sprintf("%v=%v", k, cast(tuple[1])))
+			variables = append(variables, fmt.Sprintf("%v=%v", k, cast(strings.Join(tuple[1:], "="))))
```
yesnault, line 292:
```suggestion
			variables = append(variables, fmt.Sprintf("%v=%v", k, cast(tuple[1])))
```

yesnault:
```
package main

import (
	"fmt"
	"strings"
)

func main() {
	a := "a=b=c"
	t := strings.SplitN(a, "=", 2)
	fmt.Printf("0 %v\n", t[0])
	fmt.Printf("1 %v\n", t[1])
}

```
return 

```
0 a
1 b=c

yesnault (changes requested):
I think that SplitN is more readable than `strings.Join..[1:].., '='`

cmd/venom/run/cmd.go
bramca: agreed, good suggestion

cmd/venom/run/cmd.go
bramca: Changed it to SplitN

yesnault:
LGTM, thank you

fsamin (approved):

PR #722, opened 2023-09-27
add flag VENOM_LOGS_WITH_TIMESTAMP

fokion:
When we are emitting the logs of an executor with the flag `vv` we will be overwriting with the logs of that last execution for that executor. In order to have unique logs and capture every step we need to have the timestamp in the name of the json file.

### DIFF
--- a/process_teststep.go
+++ b/process_teststep.go
@@ -61,7 +61,14 @@ func (v *Venom) RunTestStep(ctx context.Context, e ExecutorRunner, tc *TestCase,
 			if oDir == "" {
 				oDir = "."
 			}
-			filename := path.Join(oDir, fmt.Sprintf("%s.%s.step.%d.%d.dump.json", slug.Make(StringVarFromCtx(ctx, "venom.testsuite.shortName")), slug.Make(tc.Name), stepNumber, rangedIndex))
+			format := "%s.%s.step.%d.%d.dump.json"
+			name := fmt.Sprintf(format, slug.Make(StringVarFromCtx(ctx, "venom.testsuite.shortName")), slug.Make(tc.Name), stepNumber, rangedIndex)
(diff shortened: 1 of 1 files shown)


yesnault:
Hi @fokion , thank you for this proposition.

For those who read this pull-request, the issue is about that:

```
❯ ./venom run -vv b.yml
          [trac] writing venom.log
 • Hide Secret (b.yml)
        • myvar_first
                • exec PASS
                  [trac] writing b.myvar_first.step.0.0.dump.json
        • myvar_first
                • exec PASS
                  [trac] writing b.myvar_first.step.0.0.dump.json
```

Here, we have two files, with the same filename

We don't want to add a new flag about that, and perhaps that we can have a filename containing the testcase number.
As these files are linked to the venom.x.log file too, perhaps that we can add the `x` number is the filename too (and explain the dump.json filename parts in the readme).

yesnault (changes requested):
Hi @fokion , thank you for this proposition.

For those who read this pull-request, the issue is about that:

```
❯ ./venom run -vv b.yml
          [trac] writing venom.log
 • Hide Secret (b.yml)
        • myvar_first
                • exec PASS
                  [trac] writing b.myvar_first.step.0.0.dump.json
        • myvar_first
                • exec PASS
                  [trac] writing b.myvar_first.step.0.0.dump.json
```

Here, we have two files, with the same filename

We don't want to add a new flag about that, and perhaps that we can have a filename containing the testcase number.
As these files are linked to the venom.x.log file too, perhaps that we can add the `x` number is the filename too (and explain the dump.json filename parts in the readme).

ivan-velasco:
The testcase number is helpful, ideally the `testsuiteName_testcaseNumber` would make it easier to identify which log to view when errors occur.

yesnault:
see https://github.com/ovh/venom/pull/729

PR #723, opened 2023-09-27
WIP --- Update how we are loading executors 

fokion:
check them if we have them in the cache and read them only once


ovh-cds:
CDS Report build-venom-a#105.0 ✘  
* Build
  * Build ✔ 
  * Unit Tests ✔ 
* Tests
  * Acceptance Tests ✘

venom.go
```diff
@@ -231,49 +236,24 @@ func (v *Venom) registerUserExecutors(ctx context.Context, name string, vars map
...
-		var varsFromInputMap map[string]string
-		if len(varsFromInput) > 0 {
-			varsFromInputMap, err = DumpStringPreserveCase(varsFromInput)
-			if err != nil {
-				return errors.Wrapf(err, "unable to parse variables")
-			}
+		executorName, _ := getExecutorName(btes)
+		if len(executorName) == 0 {
+			fileName := filepath.Base(f)
+			executorName = strings.TrimSuffix(fileName, filepath.Ext(fileName))
```
yesnault, line 242: The name of user executor is the attribute in the file. It should not be defined by the filename.

In the read, we can see that the user executor is defined in the file lib/customA.yml, but the executor name is executor.

read_partial.go
```diff
@@ -44,6 +44,21 @@ func getVarFromPartialYML(ctx context.Context, btesIn []byte) (H, error) {
 	return partial.Vars, nil
 }
 
+func getExecutorName(btes []byte) (string, error) {
```
yesnault, line 47: Not needed, see comment below.

types_executor.go
```diff
@@ -229,12 +232,13 @@ func (v *Venom) RunUserExecutor(ctx context.Context, runner ExecutorRunner, tcIn
 		TestCaseInput: TestCaseInput{
 			Name:         ux.Executor,
 			RawTestSteps: ux.RawTestSteps,
-			Vars:         vrs,
+			Vars:         inputOnlyVrs,
```
yesnault, line 235: `tc.Vars == tc.TestCaseInput.Vars.`

I don't understand the goal of the variable inputOnlyVrs: Doing `TestCaseInput.Vars = intputOnlyVrs`, and then `tc.Vars.AddAll(vrs)` is the same as the previous code.

venom.go
```diff
@@ -216,13 +215,19 @@ func (v *Venom) getUserExecutorFilesPath(vars map[string]string) (filePaths []st
 }
 
 func (v *Venom) registerUserExecutors(ctx context.Context, name string, vars map[string]string) error {
+	_, ok := v.executorsUser[name]
```
yesnault, line 218: We can't do a cache here. The GetExecutorRunner is used in RunUserExecutor, so that `vars` can contains different value depending of the input.

yesnault (changes requested):
I'm not sure I understand the purpose of the pull-request, but if the purpose is not to read the file several times, there is already a cache on it.
If the goal is to have a cache on the executor once read with the input variables, the cache map would have to contain the input vars also in the key.

venom.go
fokion: Will mark it as WIP as I need to check if I have migrated all the necessary changes for it to work properly. 

You do need to check if you have read the executor or not as this is the point of the cache , we want to read that file only once and not multiple times ) .

PR #732, opened 2023-10-21
Increase coverage in assertions

fokion:
Adding local json instead of trying to call an http server. Increasing coverage of assertions by adding tests in the `ShouldBeMap` , `ShouldBeArray` and fixing the issue with the `ShouldNotExist`


sguiheux (approved):

assertions/assertions.go
```diff
@@ -240,10 +264,8 @@ func ShouldBeNil(actual interface{}, expected ...interface{}) error {
 
 // ShouldNotExist receives a single parameter and ensures that it is nil, blank or zero value
 func ShouldNotExist(actual interface{}, expected ...interface{}) error {
-	if ShouldBeNil(actual) != nil ||
-		ShouldBeBlank(actual) != nil ||
-		ShouldBeZeroValue(actual) != nil {
-		return fmt.Errorf("expected not exist but it was")
+	if ShouldBeNil(actual) != nil && ShouldBeBlank(actual) != nil && ShouldBeZeroValue(actual) != nil {
+		return fmt.Errorf("expected to be empty but it was not")
```
yesnault, line 268: Why do you need to update that ?

```
name: Assertions testsuite
testcases:
- name: test assertion ShouldNotExist
  steps:
  - script: echo ''
    assertions:
    - result.systemoutjson ShouldNotExist 
``` 
-->

```
ShouldBeNil: expected: Nil but is wasn't
ShouldBeBlank: <nil>
ShouldBeZeroValue: <nil> 
```

--> With your code, the testsuite is `PASS`, but should be be `FAIL`

yesnault (changes requested):

assertions/assertions.go
fokion: The documentation of this method specifies the following : 

> ShouldNotExist receives a single parameter and ensures that it is nil, blank or zero value 

I added the parameterised test that checks all these and it was failing when the value was `blank` similar to the example that you specified. This is means that the method is not working to its specification ( needs to PASS when the value is blank ) , thus the amendment

yesnault: This have to PASS (as result.systemoutjson.foo does not exist):
```yml
name: Assertions testsuite
testcases:
- name: test assertion ShouldNotExist
  steps:
  - script: echo ''
    assertions:
    - result.systemoutjson.foo ShouldNotExist
```


This have to FAIL (as result.systemoutjson exists, it's not nil):
```yml
name: Assertions testsuite
testcases:
- name: test assertion ShouldNotExist
  steps:
  - script: echo ''
    assertions:
    - result.systemoutjson ShouldNotExist
```

ShouldBeNil(actual) != nil 
if actual is NOT nil, the condition is true

If we read the whole condition: If the value is NOT nil or not blank or not zero, then something exist, so that the ShouldNotExist return an error.
NotExist is NOT something "nil AND blank AND zero"

yesnault: But yes, the condition can be simplified just by using ShouldBeNil != nil, see https://github.com/ovh/venom/commit/be5db2860edb1486cbba99fe891e7a42b158c21f

PR #735, opened 2023-10-21
Lazy expansion of json payloads that we get from user executors

fokion:
This is done in order to avoid expanding and storing big payloads that we dont need and we dont use for assertions. The flag is called `VENOM_NO_JSON_EXPANSION`.

### DIFF
--- a/process_testcase.go
+++ b/process_testcase.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
+	"reflect"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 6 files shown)


fokion:
https://github.com/ovh/venom/pull/708

process_testcase.go
```diff
@@ -574,13 +576,43 @@ func processVariableAssignments(ctx context.Context, tcName string, tcVars H, ra
 
 	for varname, assignment := range stepAssignment.Assignments {
 		Debug(ctx, "Processing %s assignment", varname)
+		lazilyLoadJSON := os.Getenv(LAZY_JSON_EXPANSION_FLAG) == FLAG_ENABLED
+		if lazilyLoadJSON {
+			if strings.Contains(assignment.From, "json") {
```
yesnault, line 581: Why do you need to check the `json` parts in the From attribute? It's possible to have a json variable without a name containing `json` string.

venom.go
```diff
@@ -27,6 +27,12 @@ var (
 	IsTest  = ""
 )
 
+const (
+	LAZY_JSON_EXPANSION_FLAG = "VENOM_NO_JSON_EXPANSION"
+	FLAG_ENABLED             = "ON"
+	FLAG_DISABLED            = "OFF"
```
yesnault, line 33: FLAG_DISABLED never used

yesnault (changes requested):
Hi @fokion, thank your for your proposition.

If I well understand, you want to cover the following usecase:
- a user executor, doing assertions on big payload
- the user executor have to be used with small payload and big payload
- The unmarshal done in the parent testcase could be long

About the implementation:
We don't want to add a global flag, unless it's a new feature that we want to test and include in the next major release.

I don't understand the move: you deleted the unmarshal from the runExector on the output by adding it on the range on the Assignments. But, if you don't want a unmarshalling of a big payload, you just need to not add the variable in the output of the user executor. 
I am aware that the unmarshal is not efficient, but I am not in favor of fixing this problem in this way.

venom.go
fokion: will delete this

process_testcase.go
fokion: we want to check the assignments for the keys that we are generating from the unmarshalling. For example we are making new keys with the name `result.systemoutjson` for the key `result.systemout` the check here is to check if the  key has that `json` substring in order to check if we need to unmarshal that payload.

types_executor.go
```diff
@@ -164,6 +164,26 @@ func GetExecutorResult(r interface{}) map[string]interface{} {
...
+		for key, value := range d {
+			switch z := value.(type) {
+			case string:
+				m, e := UnmarshalJSON(key, z)
+				if e != nil {
+					panic(e)
+				}
+				if len(m) > 0 {
+					for s, i := range m {
+						if !strings.Contains(strings.ToUpper(s), "__Type__") && !strings.Contains(strings.ToUpper(s), "__Len__") {
```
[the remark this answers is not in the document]
fokion: thanks will need to fix this.

fokion:
The idea is that you can have a big JSON payload , by default we will try to unmarshal it and we may not use the new keys in any assertion. That will cause issues with performance as the constant unmarshalling , copying of values with increase memory and cpu usage. The point here is to unmarshal this payload lazily and for now have that feature tested under a flag so that we dont affect everyone.

yesnault:
Closing this pull-request for now: we don't want to add a global flag, unless it's a new feature that we want to test and include in the next major release.

PR #767, opened 2024-02-19
exec: add a 'stdin' attribute

cdevienne:
If set, the value is written to the script stdin. This avoid the need to use 'echo' in the script, which makes thing clearer.

It is also particularly useful in the following case (probably other issues with variable substitution):

```yaml
name: simple_test
vars:
  payload: {"s": "\n"}
testcases:
- name: simple_test
  steps:
  - type: exec
    script: echo {{.payload | toJSON}} | jq .
    assertions:
    - result.code MustEqual 0
```

This will fail because the \n is not properly escaped.

Using the new stdin attribute solves the issue:

```yaml
name: simple_test
vars:
  payload: {"s": "\n"}
testcases:
- name: simple_test
  steps:
  - type: exec
    stdin: {{.payload | toJSON}}
    script: jq .
    assertions:
    - result.code MustEqual 0
```

### DIFF
--- a/executors/exec/README.md
+++ b/executors/exec/README.md
@@ -34,6 +34,18 @@ testcases:
             echo "Bar"
 ```
 
+Input:
+
+```yaml
+name: Title of TestSuite
+testcases:
+- name: with stdin
+  steps:
+  - type: exec
+    stdin: Foo
+    script: cat
+```
+
 ## Output
 
 ```yaml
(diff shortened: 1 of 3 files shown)


yesnault (approved):

ovh-cds:
CDS Report build-venom-a#173.0 ✘  
* Build
  * Build ✔ 
  * Unit Tests ✔ 
* Tests
  * Acceptance Tests ✘

yesnault (approved):

PR #768, opened 2024-02-20
 Add a 'command' property to 'exec'

cdevienne:
Add a 'command' property to 'exec'

It is an alternative to 'script' that takes an explicit list of strings
and does not use a shell.

It makes it easier to deal with escaping special characters, and gives
more control on how the command is run.

:warning: It is based on PR #767


yesnault:
https://github.com/ovh/venom/pull/782

PR #792, opened 2024-05-09
fix: [Make] Create windows binaries with .exe extension

zalgonoise:
When fetching Venom binaries, I noticed that Windows releases didn't include an extension. For Windows, it means that the users will face the program-picker prompt due to a lack of extension, and would have to add it manually. 

This PR changes the `CROSS_COMPILED_BINARIES` make target in `.build/go.mk` to check if the current target OS is windows, and if so, changing the filename to include the `.exe` extension.

I assume that the point of `BINARIES` is precisely this, but probably due to some lack of Make syntax knowledge, I couldn't figure out why `IS_WINDOWS` isn't being recognized :) 

That being the case, I am opening a suggestion to leverage the `GOOS` env var taken from the `get_os_from_binary_file` call, and changing the filename if the OS is `windows`.

Any suggestions for changes are welcome, provided that we are able to distribute windows binaries with the proper extension :)

### DIFF
--- a/.build/go.mk
+++ b/.build/go.mk
@@ -48,9 +48,12 @@ mk_go_build_clean:
 
 $(CROSS_COMPILED_BINARIES): $(GOFILES) $(TARGET_DIST)
 	$(info *** compiling $@)
-	@GOOS=$(call get_os_from_binary_file,$@) \
+	@os=$(call get_os_from_binary_file,$@); \
+	filename=$@ ; \
+	if [[ $${os} == 'windows' ]]; then filename=$@.exe; fi; \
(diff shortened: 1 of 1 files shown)


zalgonoise:
Note: the force-push was to add the sign-off mark to the commit, in compliance with the DCO :)

yesnault:
Thank you @zalgonoise

ovh-cds:
CDS Report build-venom-a#190.0 ✘  
* Build
  * Build ✔ 
  * Unit Tests ✔ 
* Tests
  * Acceptance Tests ✘

PR #798, opened 2024-06-11
feat(executor/rabbitmq): Add support for RPC

docknight:
# Description
Introduced new client type to use for testing RPC pattern using [Direct Reply-to](https://www.rabbitmq.com/docs/direct-reply-to).
This can not be achieved using existing client types (publisher and subscriber), because a Direct reply-to has some specific requirements and the client must have a reply consumer running before publishing a request.
The new client type is simply referred to as 'client' to indicate a client-server communication using RPC pattern.
It can be used to test a server that is expected to respond to the received message by sending a reply using Direct Reply-to feature of RabbitMQ.

# Usage
Example testsuite:
```
name: TestSuite RabbitMQ
testcases:
  - name: Send data and await reply
    steps:
      - type: rabbitmq
        addrs: "{{.addrs}}"
        user: "{{.user}}"
        password: "{{.password}}"
        clientType: client
        exchange: test-exchange
        exchangeType: direct
        durable: true
        routingKey: OrderDetailsRequest
        messages: 
          - value: '{{.messageFromVariableFile}}'
            contentType: application/json
            contentEncoding: utf8
            persistant: false
            durable: true
        assertions: 
          - result.bodyjson.bodyjson0 ShouldContainKey Status
          - result.bodyjson.bodyjson0.Status ShouldEqual Ok
```
In the above example, a direct-reply-to consumer will first be started and then a message will be published to the exchange using the provided routing key. Then, the test executor will wait for the reply to be received on the direct reply queue. Existing logic of the subscriber client type will be used to parse the results, close the connection and validate against assertions.

### DIFF
--- a/executors/rabbitmq/README.md
+++ b/executors/rabbitmq/README.md
@@ -1,6 +1,11 @@
 # Venom - Executor RabbitMQ
 
-Step to use publish / subscribe on a RabbitMQ
+Three types of execution are supported:
+- **publisher**: publish a message to a queue or to an exchange.
+- **subscriber**: bind to a queue or an exchange (using routing key) and wait for message(s) to be consumed.
+- **client**: publish a message to a queue or to an exchange and wait for the response message to be received on the [reply-to](https://www.rabbitmq.com/docs/direct-reply-to) queue.
+
+Steps to use publish / subscribe on a RabbitMQ:
 
 ## Input
 In your yaml file, you can use:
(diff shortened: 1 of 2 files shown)


yesnault (changes requested):
Thank you @docknight 

Can you provide some documentation about that into 
https://github.com/ovh/venom/blob/master/executors/rabbitmq/README.md please?

docknight:
@yesnault I've extended the documentation with an introduction of supported execution types and an example of using the RPC execution type in a testsuite.

yesnault (approved):

PR #809, opened 2024-08-28
fix(rabbitmq-rpc): Fix channel initialization for message publishing.

docknight:
Recent feature (RPC calls for rabbitmq executor) broke the publishing type of execution. 
When running the testsuite, execution faulted with the following error:
"rabbitmqpanic: runtime error: invalid memory address or nil pointer dereference",

This has been fixed and tested. A testsuite with a client type set to 'publisher' does not throw a nil pointer error and is properly publishing message(s) to the exchange.


ovh-cds:
CDS Report build-venom-a#208.0 ✘  
* Build
  * Build ✘  
  * Unit Tests ✘

yesnault:
@docknight 

../../executors/rabbitmq/rabbitmq.go:161:19: cannot use e.openChannel(ctx) (value of type *amqp.Connection) as *amqp.Channel value in assignment
29s996ms
../../executors/rabbitmq/rabbitmq.go:161:19: cannot use e.openChannel(ctx) (value of type *amqp.Channel) as *amqp.Connection value in assignment

yesnault (approved):
Thank you @docknight

PR #821, opened 2024-11-03
feat: add NATS executor

### LINKED ISSUE #820: feature: NATS executor
Hi, 

it would be nice to have a NATS executor.

LuBashQ:
Add NATS executor, which allows to test services that use NATS as their communication layer.

Currently only publish and subscribe are supported, both with and without Jetstream, but the plan is to also add [KeyValue](https://docs.nats.io/nats-concepts/jetstream/key-value-store) commands (e.g "check if a specific value is in a KeyValue store" tests).

Closes #820


ovh-cds:
CDS Report build-venom-a#257.0 ✘  
* Build
  * Build ✔ 
  * Unit Tests ✔ 
* Tests
  * Acceptance Tests ✘

yesnault (changes requested):
Thank your @LuBashQ for this new executor.

The binary `nats` is not available into the integration tests. Can you execute the nats commands from a container nats started with integration tests Makefile ?

PR #843, opened 2025-04-20
Feature: Playwright executor

zikani03:
This PR adds a [Playwright](https://playwright.dev/) support to venom to allow users to perform end-to-end/integration Browser tests using Playwright. It uses the playwright-go library to perform the steps.

Steps to execute actions within playwright are called "Actions" and we have several actions supported so far (see the readme). 

Some elements that may need further discussion before merging (if this feature is approved) include
- Configuration and handling of timeouts 
- Ability to pass options to different actions (currently most actions use the default options playwright-go uses)
- Whether to support running tests with multiple browsers (instead of just one)

## Example usage

This PR enables the following

```yaml
name: Playwright testsuite
testcases:
- name: Check the title
  steps:
    - type: playwright
      url: https://localhost:5173
      headless: true
      actions:
        - Fill "#email" "change@example.com"
        - Fill "#email" "zikani@example.com"
        - Fill "#password" "password"
        - Click "#loginButton"
        - WaitFor ".welcome-dashboard"
      assertions:
        - result.page.body ShouldContainSubstring Test Application
        - result.document.body ShouldContainSubstring Hello,&nbsp;Zikani
        - result.document.body ShouldContainSubstring Logout
```


zikani03:
Hello @yesnault ,

I think that approach is fine. Though, writing and reading the actions as one string felt like it would be faster and more human-friendly. I don't know if there is a way we can strike a balance between typing structured YAML and that shorter syntax.

Besides that, I would just suggest that the action have the keys named differently, I'd suggest the following

```yaml
actions:
  - action: fill
    selector: "#email"
    content: "change@example.com"
```

zikani03:
@yesnault I have pushed some changes and settled for the more structured approach for now. I created a [separate repo](https://github.com/zikani03/pact) to explore the DSL but I think for this PR the more structured approach works better as it requires less explanation to users.
I am also comfortable with it since those who want to keep the file compact, like me, can always write the actions in one line like this :)

```yaml
- { action: Fill, selector: "#password", content: "password" }
```

zikani03:
Playwright options for the actions are now supported

```yaml
    actions:
      - action: Fill
        selector: "#email"
        content: "johndoe@example.com"
        options:
          timeout: 60000
```

PR #856, opened 2025-05-28
fix(reports): use filepath.Join instead or linux sep or path.Join for file that will be written on OS - fixes #794

### LINKED ISSUE #794: Output file name is pre-pended with path to test file (Windows version)
The path to the yaml file is being pre-prended onto the output file.

When I run the test with the yaml in a non-local folder:

```
venom.windows-amd64.exe run  \Users\joe\Downloads\tests_suite.yml --var="target_site=https://xyz.com"   --format=json --output-dir=venom-test
```

I get this error (path misconstruction):

```
Error while creating file venom-test/test_results_\Users\joe\Downloads\tests_suite.json: open venom-test/test_results_\Users\joe\Downloads\tests_suite.json: The system cannot find the path specified.
```

Is this expected?

kilianpaquier:
Fixes an issue for Windows where report files can't be written since we were replacing / by _ and not taking into account Windows separator \.

Proposals:
1. Only use testfile name and not the whole path for the test result filename (current implementation).
2. Extend replace to os.PathSeparator which will replace \ or / depending on executing OS.

### DIFF
--- a/venom_output.go
+++ b/venom_output.go
@@ -7,7 +7,6 @@ import (
 	"encoding/xml"
 	"fmt"
 	"os"
-	"path"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


venom_output.go
```diff
@@ -109,9 +108,8 @@ func (v *Venom) OutputResult() error {
 			return errors.New("Error: you have to use the --html-report flag")
 		}
 
-		fname := strings.TrimSuffix(ts.Filepath, filepath.Ext(ts.Filepath))
-		fname = strings.ReplaceAll(fname, "/", "_")
```
kilianpaquier, line 113: **info:** Real fix to do is here where on windows it is `\` and not `/`. We could use `os.PathSeparator` as stated in PR description but I think we could just keep only the testfile for result filename

yesnault (approved):
LGTM

PR #863, opened 2025-06-12
Feature/issue 858

### LINKED ISSUE #858: Add ShouldNotJSONEqual
Hi!

I think it would be super useful to have the negation of `ShouldJSONEqual` as `ShouldNotJSONEqual`.

I need to check if a JSON has changed from one step to the next one, and as it is now I can only check the Equality. I think is easy enough to implement, and it can be very useful (same as the other negations ShouldNotJSONContain or ShouldNotJSONContainWithKey).

Thanks in advance :)

09bassOp:
Closes #858 

New assertion: `ShouldNotJSONEqual`


pacomoreno-bandai (approved):
👏

yesnault:
Thank you @09bassOp

PR #864, opened 2025-06-12
feat(http): support cookies in redirects

Crocmagnon:
This PR aims at supporting cookies in same-step redirects.
It does not handle cookies more globally (across steps).


Crocmagnon:
See #866

PR #865, opened 2025-06-12
fix(user-executors): ensure unique lib paths in user executor files retrieval

ivan-velasco:
- Prevent "executor already exists" error caused by same directory being walked twice with different path representations (absolute vs relative)


ivan-velasco:
If `v.LibDir` is set to an absolute path that is the same as the relative path (e.g., `/home/user/project/lib` and `venom.testsuite.workdir` is `/home/user/project`), both `/home/user/project/lib` (absolute) and `/home/user/project/lib` (relative) are added to `libpaths`. This results in the same directory being walked twice, but with different path representations (absolute vs. relative). When collecting file paths, the map `filepaths` uses the full path as the key, so the same file can be added twice under different keys if the directory is referenced both ways.

PR #899, opened 2026-02-08
feat(spanner): Add Google Spanner executor

picone:
## Summary
This PR adds a new `spanner` executor to run SQL statements against Google Cloud Spanner from Venom test suites.

## What’s included
- New step type: `type: spanner`
- Inputs:
  - `project`, `instance`, `database` (required)
  - `credentials_file` (optional; defaults to Application Default Credentials)
  - `commands` (list of SQL statements) or `file` (fallback when `commands` is empty)
- Execution:
  - Statements run sequentially
  - DML is detected by leading keyword (`INSERT`/`UPDATE`/`DELETE`) and executed in a read-write transaction
  - Other statements are treated as queries
- Output:
  - Per statement result in `result.queries[]` with:
    - `statement_type: query` + `rows: []`
    - `statement_type: dml` + `row_count: <number>`

## Example
```yaml
- type: spanner
  project: my-project
  instance: my-instance
  database: my-database
  commands:
    - "SELECT SingerId, FirstName FROM Singers LIMIT 1"
    - "UPDATE Singers SET FirstName = 'Alice' WHERE SingerId = 1"

### CODE REVIEW

**Comment by yesnault** [comment]
Hi @picone , thanks a lot for this proposal.

After reviewing, we've decided not to merge this PR as a native Go executor. The reasoning, to be transparent:

1. go.mod cost. cloud.google.com/go/spanner pulls a large transitive chain (gRPC, protobuf, OpenTelemetry, …)
2. This is exactly what user-executors are for. When a feature can be expressed by wrapping an existing CLI, our preference is to keep the venom binary lean and let users compose with a YAML user-executor. We want to be conservative about
  adding new native executors — each one is a long-term maintenance commitment (dep bumps, CVE tracking, emulator-based tests…), and that bar is harder to clear when a CLI already covers the use case cleanly.

This would also be a good showcase of what user-executors can do — wrapping a CLI is exactly the pattern they were designed for.

  Proposed lib/spanner.yml
```yml
executor: spanner
input:
project: ""
instance: ""
database: ""
sql: ""
credentials_file: ""
format: "json"
steps:
- type: exec
script: |
    {{- if ne .input.credentials_file ""}}
    export GOOGLE_APPLICATION_CREDENTIALS="{{.input.credentials_file}}"
    {{- end}}
    gcloud spanner databases execute-sql "{{.input.database}}" \
    --project="{{.input.project}}" \
    --instance="{{.input.instance}}" \
    --sql="{{.input.sql}}" \
    --format="{{.input.format}}"
assertions:
- result.code ShouldEqual 0
vars:
    raw:
    from: result.systemout
    payload:
    from: result.systemoutjson
output:
raw: "{{.raw}}"
payload: "{{.payload}}"

```

Usage in a testsuite
```yml
name: Spanner sample
testcases:
- name: query
steps:
- type: spanner
    project: my-project
    instance: my-instance
    database: my-db
    sql: "SELECT SingerId, FirstName FROM Singers LIMIT 1"
    assertions:
    - result.payload.rows.__Len__ ShouldEqual 1

- name: dml update
steps:
- type: spanner
    project: my-project
    instance: my-instance
    database: my-db
    sql: "UPDATE Singers SET FirstName = 'Alice' WHERE SingerId = 1"
    credentials_file: /path/to/sa.json
    assertions:
    - result.code ShouldEqual 0
```

Run with `venom run testsuite.yml --lib-dir ./lib`

Trade-offs to be transparent about

- Runtime dependency on gcloud SDK instead of a self-contained binary. Acceptable IMO for a Spanner-targeting test environment, since users typically already have gcloud set up to authenticate ADC.
- The JSON shape comes from gcloud (metadata.rowType.fields[] + rows[]) rather than your custom result.queries[]. Different, but documented — and gives users the full Spanner metadata for free.
- Multi-statement atomic transactions aren't exposed, but the current PR doesn't expose them either (each DML in commands runs in its own ReadWriteTransaction).

PR #902, opened 2026-02-16
fix(executors/mongo): Use SetProjection in projection option

jeanschmitt:
Hi,

This PR fixes the mongo option used by the Projection option in mongo executor.


yesnault:
Thank you @jeanschmitt

PR #907, opened 2026-03-25
Fix user executor failures missing from JUnit/XML/JSON/HTML reports

ivan-velasco:
**Problem:** User-executor inner failures were only merged into the parent step inside printTestStepResult, so paths that skipped that call left reports without real error detail.

**Change:** After runTestSteps in RunUserExecutor, merge all inner TestStepResults into the parent (Errors, ComputedInfo, stdout/stderr; skip whitespace-only streams). Drop the merge from printTestStepResult to avoid duplicates.

**Tests:** tests/user_executor_with_failure.yml plus lib executors (simple + nested failure).

example
```xml
<testcase classname="user_executor_with_failure.yml" name="simple-user-executor-failure" time="0.014">
  <error><![CDATA[Testcase "user_executor_with_failure", step #2-0: Assertion "result.code ShouldEqual 1" failed. expected: 1  got: 0 (user_executor_with_failure.yml:0)]]></error>
  <error><![CDATA[Testcase "simple user executor failure", step #1-0: executor "user_executor_with_failure" failed (user_executor_with_failure.yml:0)]]></error>
  <system-out><![CDATA[this step succeeds
failing step
]]></system-out>
</testcase>
```


yesnault:
thank you @ivan-velasco
see https://github.com/ovh/venom/pull/912 for next

PR #913, opened 2026-04-27
fix(venom): some hardening

yesnault:
HTML Report more compact / add filter and search.

TLS, SSH host key, path traversal, secret logging
  - Drop the loop logging secret names on testsuite startup.
  - Add ResolveWorkdirPath helper rejecting absolute paths and ".." escapes;
    apply it across 8 file-based executors (sql, dbfixtures, redis, mongo,
    kafka, ovhapi, http, readfile).
  - SMTP/IMAP: TLS now verified by default; opt-out via `ignore_verify_ssl`.
  - SSH: host key verified against ~/.ssh/known_hosts by default; opt-out via
    `insecure_ignore_host_key`. Removed os.ExpandEnv on `privatekey`.
  - Update tests/ssh.yml accordingly.

Plugin loading, HTTP TLS paths, exec hardening, SSH timeout,
  HTML XSS, SSH privatekey
  - venom.go: load .so plugins only from --lib-dir (no auto-discovery from
    workdir); validate plugin name against [A-Za-z0-9_-].
  - HTTP: reject absolute paths for tls_root_ca / tls_client_cert /
    tls_client_key; detect inline PEM via "-----BEGIN".
  - exec: create temp script atomically (O_EXCL, 0o700); fix Windows .ps1
    rename; document stdout/stderr synchronization.
  - SSH: add `timeout` field (default 30s); restrict privatekey absolute
    paths to $HOME, resolve relative paths under workdir.
  - HTML report: sanitize markdown rendering with DOMPurify.
  - Update READMEs (http, smtp, imap, ssh, plugins).

Dependencies, update checksums, DSN redaction
  - Migrate executors/rabbitmq from archived streadway/amqp to
    rabbitmq/amqp091-go (drop-in).
  - Remove direct dependency on mattn/go-sqlite3; consolidate on
    modernc.org/sqlite with `sqlite3` alias for YAML backward compatibility.
  - venom update: download checksums.txt from the GitHub release and verify
    SHA256 of the binary before applying; refuse update if checksums absent.
  - Add RedactURI helper and apply it to SQL, dbfixtures and Mongo
    connection logs to prevent credential leaks via venom.log.
  - Fix missing `mapstructure:"..."` tags on snake_case bool/int fields
    (ignore_verify_ssl, insecure_ignore_host_key, timeout) which silently
    reverted the opt-in/opt-out flags to zero-value.

  BREAKING CHANGES
  - SMTP/IMAP require `ignore_verify_ssl: true` for self-signed certs.
  - SSH requires populated ~/.ssh/known_hosts or
    `insecure_ignore_host_key: true`.
  - File paths in the affected executors must be relative to the testsuite
    workdir; absolute paths are rejected.
  - HTTP TLS file paths must be relative to workdir.
  - SSH privatekey absolute paths must be under $HOME.
  - .so plugins now require --lib-dir.
  - `venom update` requires the GitHub release to publish checksums.txt
    alongside binaries.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,11 @@ $(ALL_RESULTS_TARGETS):
 	$(info copying $(call get_results_from_target, $@) to $@)
 	@cp -f $(call get_results_from_target, $@) $@
 
-.PHONY: build lint clean testrun test dist test-xunit package
+.PHONY: build lint clean testrun test dist test-xunit package vulncheck checksums
+
+vulncheck: ## scan dependencies for known CVEs (uses golang.org/x/vuln/cmd/govulncheck)
+	@which govulncheck > /dev/null 2>&1 || go install golang.org/x/vuln/cmd/govulncheck@latest
+	govulncheck ./...
 
 build: ## build all components and push them into dist directory
 	$(info Building Component venom)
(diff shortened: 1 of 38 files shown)


yesnault:
example: 
Screenshot 2026-04-27 at 23 52 37

PR #914, opened 2026-06-03
feat: add test suite and test case parallelization

floriancazals:
Summary
This PR adds two levels of parallelization to Venom test execution and fixes a bug that prevented the existing intra-testsuite parallelization from working.

Changes
Bug fix: [parallel](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) field was never propagated from YAML to runtime

The [Parallel](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) field on TestSuiteInput was parsed from YAML but never copied into the TestSuite struct used at runtime, making the existing worker pool code unreachable.

New: cross-testcase dependency detection

Before running testcases in parallel, Venom now scans steps for references like [{{.otherTestcase.variable}}](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html). If a dependency is detected, it automatically falls back to sequential execution with a warning — preventing subtle race conditions where a testcase would read variables not yet produced by another.

New: [--parallel-suites](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html) flag for inter-suite parallelism

Allows running multiple .yml testsuite files concurrently via a bounded worker pool. Configurable via:

CLI: [--parallel-suites=N](vscode-file://vscode-app/usr/share/code/resources/app/out/vs/code/electron-browser/workbench/workbench.html)
Env: VENOM_PARALLEL_SUITES=N
Config file (.venomrc): parallel_suites: N
Output is buffered per-suite to prevent interleaving.

Usage
```
# Intra-suite: run testcases in parallel within a single file
name: My Suite
parallel: 4
testcases: [...]
```

```
# Inter-suite: run multiple test files in parallel
venom run ./tests/ --parallel-suites=4
```

### DIFF
--- a/cmd/venom/run/cmd.go
+++ b/cmd/venom/run/cmd.go
@@ -25,24 +25,26 @@ var (
 	path []string
 	v    *venom.Venom
 
-	variables     []string
-	secrets       []string
-	format        string = "xml" // Set the default value for formatFlag
-	varFiles      []string
-	outputDir     string
-	libDir        string
-	htmlReport    bool
-	stopOnFailure bool
-	verbose       int = 0 // Set the default value for verboseFlag
+	variables      []string
+	secrets        []string
+	format         string = "xml" // Set the default value for formatFlag
+	varFiles       []string
+	outputDir      string
+	libDir         string
+	htmlReport     bool
+	stopOnFailure  bool
+	verbose        int = 0 // Set the default value for verboseFlag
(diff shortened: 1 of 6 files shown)


yesnault:
this feature can be done with parallel

ls *.yml|parallel -j 2 venom run {}

PR #917, opened 2026-06-06
fix: Use DumpStringPreserveCase for computedVars in RunUserExecutor

ivan-velasco:
User-executor output fails to interpolate camelCase step vars (e.g. `{{.staticVersionLocator}}`)

**Step vars:** assignment succeeds, but executor output leaves camelCase templates as literal strings. Snake_case/lowercase vars work.

**Cause:** RunUserExecutor dumps computedVars with DumpString (lowercases keys), then `interpolate.Do` does case-sensitive unknown-var detection. staticVersionLocator ≠ staticversionlocator → treated as unknown → literal preserved.

**Fix:** Use DumpStringPreserveCase for user-executor output interpolation (consistent with step interpolation).

**Workaround:** Use snake_case var names.


yesnault:
so many commits on this branch, rework done on https://github.com/ovh/venom/pull/934

PR #936, opened 2026-09-02
(feat):Enhance multipart file handling and support for repeated fields

ivan-velasco:
Adds two related improvements to multipart form handling in the HTTP executor:                                                                                                                           
  - Content-Type override: file uploads via `@path` can now specify `@path;type=content/type` to set the part's Content-Type, instead of always getting the hardcoded `application/octet-stream`.   
  - Repeated fields: multipart_form values can now be a list of strings for multi-file uploads.    

**Example**
```yaml
- name: upload a file to delete
  type: http
  method: POST
  url: "https://api.example.com/route/test-me?type=builder"
  headers:
    Authorization: "Bearer dummy_token_1234"
    X-Dummy-Header: "example-value"
  multipart_form:
    files: "@testdata/dummyfile.txt;type=text/plain"
    description: "dummy upload file for sharing example"
  assertions:
    - result.statuscode ShouldEqual 201
    - result.bodyjson.files.files0.path ShouldEqual "/dummy/path/file.txt"
  vars:
    uploaded_path:
      from: result.bodyjson.files.files0.path
    uploaded_file_size:
      from: result.bodyjson.files.files0.size
```

### DIFF
--- a/executors/http/http.go
+++ b/executors/http/http.go
@@ -12,6 +12,7 @@ import (
 	"net"
 	"net/http"
 	"net/http/cookiejar"
+	"net/textproto"
 	"net/url"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


executors/http/http.go
```diff
@@ -377,6 +380,44 @@ func (e Executor) getRequest(ctx context.Context, workdir string) (*http.Request
...
+		// multipart.Writer.CreateFormFile hardcodes application/octet-stream
+		// and can't be overridden, which some servers reject via Content-Type
+		// allow-lists.
+		if idx := strings.LastIndex(filePath, ";type="); idx != -1 {
+			contentType = filePath[idx+len(";type="):]
+			filePath = filePath[:idx]
+		}
+		if _, err := os.Stat(filePath); !os.IsNotExist(err) {
+			header := textproto.MIMEHeader{}
+			header.Set("Content-Disposition", fmt.Sprintf(`form-data; name="%s"; filename="%s"`, escapeQuotes(key), escapeQuotes(filepath.Base(filePath))))
```
yesnault, line 409: Should ne replaced by: 

```
header.Set("Content-Disposition", multipart.FileContentDisposition(key, filepath.Base(filePath)))
```

And no need to use espaceQuotes probably

yesnault: Can you add something in the Readme (http executor too), something like that:

```
## Multipart form

 `multipart_form` sends the request body as `multipart/form-data`. It cannot be combined with `body` or `bodyFile`.

 Each key is a form field. A value is either a single string, or a list of strings
 to send the same field name once per item (e.g. a repeated `files` part for a multi-file upload).

 A value starting with `@` is the path of a file to upload, resolved from the current
 working directory. Append `;type=<content-type>` to set the Content-Type of that part;
 it defaults to `application/octet-stream`. This mirrors curl's `-F field=@path;type=...`.

 ```yaml
 - name: upload files
   steps:
   - type: http
     method: POST
     url: https://httpbin.org/post
     multipart_form:
       description: "a plain text field"
       tags:
         - "first"
         - "second"
       files:
         - "@./testdata/report.pdf;type=application/pdf"
         - "@./testdata/notes.txt;type=text/plain"
     assertions:
 ```

yesnault (changes requested):

yesnault (approved):

yesnault:
Thank you @ivan-velasco

